PR #7, opened 2019-07-05
Add bolt db package

wParkhi:


jyong2kim:
@wParkhi 
Please resolve conflict.

ChaJiWon:
@wParkhi,
Upper conflicts is related with my commits, so I resolved it.

src/db/bolt/network/network.go
```diff
@@ -0,0 +1,162 @@
...
+	Wired map[string]string    `json:"wired"`
+	Wireless map[string]string `json:"wireless"`
+}
+
+type NetworkInfo struct {
+	Id       string `json:"id"`
+	IPv4 Addr       `json:"IPv4"`
+	RTT float64     `json:"RTT"`
+	// IPv4Wired map[string]string `json:"IPv4Wired"`
+	// IPv4Wireless map[string]string `json:"IPv4Wireless"`
```
jyong2kim: If this commented line is not used, please remove.

src/common/errors/errors.go
```diff
@@ -15,6 +15,7 @@
  *
  *******************************************************************************/
 
+// Package errors defines error cases of edge-orchestration.
 package errors
```
jyong2kim: How about to change package name?
because this package name can be confused with golang default package "errors".

src/db/bolt/network/network.go
wParkhi: okay

src/common/errors/errors.go
wParkhi: I agree with the suggestion that there will be confusion due to the package name. After the discussion, I will reflect it as another PR.

glide.yaml
```diff
@@ -94,3 +100,4 @@ import:
     version: v1.4.0
   - package: github.com/spf13/pflag
     version: v1.0.3
+  - package: github.com/boltdb/bolt
```
wansuyoo: It is better to describe the exact release version.
Currently, latest release is v1.3.1

jyong2kim: it is added.

mgjeong:
Looks good to me.

PR #15, opened 2019-07-22
Changed scoring flow to include RTT

wParkhi:
- Check RTT for each device and store it in DB.
- Check the resource (cpu, memory, network) in the device for every reference time (now, 5sec) and store it in db.
- getScore returns the score from the information stored in the DB when requested


wParkhi:
@mgjeong  please review this pull request :)

mgjeong:
Looks good.. thanks..

PR #18, opened 2019-08-05
Update RequestService API for execution service to multiple platform

jyong2kim:


src/orchestrationapi/orchestration_api.go
```diff
@@ -61,53 +65,131 @@ type orcheClient struct {
...
+	ServiceInfo []RequestServiceInfo
+	// TODO add status callback
+}
+
+type TargetInfo struct {
+	ExecutionType string
+	Target        string
+}
+
+type ReponseService struct {
```
wansuyoo: typo?
*Response

jyong2kim: Thanks. I will fix.

PR #28, opened 2019-08-20
Apply external api changes

wansuyoo:
-. Update interface(C/Java) of go library according to changes of external API.
-. Add db at GoMain for docker container platform.
-. Fix http persistent connection issue.


GoMain/src/main/main.go
```diff
@@ -80,23 +80,23 @@ func main() {
...
-		serviceInfo[0].ExecutionType = "container"
-		serviceInfo[0].ExeCmd = []string{"docker", "run", "-v", "/var/run:/var/run:rw", "hello-world"}
-		requestService := orchestrationapi.ReqeustService{
-			ServiceName: "container_service",
-			ServiceInfo: serviceInfo,
-		}
-
-		externalapi.RequestService(requestService)
-	}
+	// externalapi, err := orchestrationapi.GetExternalAPI()
```
jyong2kim: If it is no needed, Please remove it.
and if it was for guidance for use, insert comments

wansuyoo: Yes, i think it's better to remove it.
Guide of rest-api is in README.

PR #75, opened 2019-12-03
update the android service package name

wansuyoo:
Package name we kept has com.samsung.homeedge , since "-" is not allowed in android

### DIFF
--- a/src/interfaces/javaapi/javaapi_secured.go
+++ b/src/interfaces/javaapi/javaapi_secured.go
@@ -116,7 +116,7 @@ const (
 	platform      = "android"
 	executionType = "android"
 
-	edgeDir = "/data/user/0/com.samsung.orchestration.service/files"
+	edgeDir = "/data/user/0/com.samsung.homeedge/files"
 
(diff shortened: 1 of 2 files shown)


src/interfaces/javaapi/javaapi_secured.go
```diff
@@ -116,7 +116,7 @@ const (
 	platform      = "android"
 	executionType = "android"
 
-	edgeDir = "/data/user/0/com.samsung.orchestration.service/files"
+	edgeDir = "/data/user/0/com.samsung.homeedge/files"
```
jyong2kim: How about moving this variable to input parameter on javaapi?

wansuyoo: Yes, as we discussed about this issue, I think it's better to separate this value from edge-orchestration module.
It doesn't look good to place this constant value in orchestration module.
I'll check and prepare with better method.

jyong2kim: Thanks :+1:

PR #84, opened 2020-03-25
Add alternative way to get processor frequency (/proc/cpuinfo).

### LINKED ISSUE #81: in kvm machine , it fail to pass init
**Describe the bug**
AS title shows 

**To Reproduce**

set up docker env on kvm based VM , and get key , user id ready 
then launch it 

**Expected behavior**
get cpu info step pass , 

**Screenshots**
2020/03/10 08:54:04 cpu.go:100: resourceutil cpu info getting fail :  open /sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq: no such file or director

**Test environment configuration (please complete the following information):**
root@ubuntu18:/var/edge-orchestration/log# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic
root@ubuntu18:/var/edge-orchestration/log# uname -a
Linux ubuntu18 4.15.0-88-generic #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
root@ubuntu18:/var/edge-orchestration/log# virt-what
virtualbox
kvm

* Firmware version: (OS type and version, e.g., Ubuntu 14.04, etc.)
* Hardware: (e.g., x86-64, ARM, CPU type, RAM size, etc.)
* Toolchain: (e.g., Docker and Go versions)
* Edge Orchestration Release: (e.g., Alpha, Baobab, Coconut)

**Additional context**

tdrozdovsky:

# Description

An alternative method was added to obtain the processor frequency for calculating performance score (the value is read from the  `/proc/cpuinfo` file). 

By default, the frequency value is read from `/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq`, but in case of a failure it reads from the `/proc/cpuinfo` file (this happens when Linux starts on a Virtual Machine), the default `/proc/cpuinfo` cannot be used, since the current frequency displayed in the `/proc/cpuinfo` may differ from the maximum (the processor may be in low power mode).

Fixes #81

## Type of change

# How Has This Been Tested?

The Edge Orchestration must run successfully in a Virtual Machine and calculate own performance score.

**Test Configuration**:
* Firmware version: **Host**: Windows 10 - VirtualBox **Guest**: Ubuntu 16.04
* Hardware: x86-64
* Toolchain: recommended by default Docker and Go versions

# Checklist:

### DIFF
--- a/src/common/resourceutil/cpu/cpu.go
+++ b/src/common/resourceutil/cpu/cpu.go
@@ -116,7 +116,10 @@ func Info() ([]InfoStat, error) {
 
 	feq, err := getCPUMaxFreq()
 	if err != nil {
-		return nil, err
+		feq, err = getCPUFreqCpuInfo()
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	ret[0].Mhz = feq
@@ -149,6 +152,51 @@ func getCPUMaxFreq() (float64, error) {
 	}
 
 	return feq, nil
+}
+
+func getCPUFreqCpuInfo() (float64, error) {
+	f, err := fileOpen("/proc/cpuinfo")
+	defer f.Close()
+	if err != nil {
+		return 0.0, err
+	}
+
+	r := bufio.NewReader(f)
+	for {
+		line, err := r.ReadString('\n')
+		if err != nil {
+			break
+		}
+
+		fields := strings.Split(strings.Trim(line, "\n"), ":")
+		if len(fields) < 2 {
+			continue
+		}
+
+		switch strings.TrimSpace(fields[0]) {
+		//case "cpu MHz":
+		//	feq, err := strconv.ParseFloat(strings.TrimSpace(fields[1]), 64)
+		//	if err != nil {
+		//		return 0.0, err
+		//	}
+		//	return feq, nil
+		case "model name":
+			name := strings.Split(fields[1], " ")
+
+			for _, val := range name {
+				if strings.Contains(val, "GHz") {
+					feq, err := strconv.ParseFloat(strings.Trim(val, "GHz"), 64)
+					if err != nil {
+							return 0.0, err
+						}
+					feq *= 1000.0
+					return feq, nil
+				}
+			}
+		}
+	}
+
+	return 0.0, err
 }
 
 func getCPUs() ([]InfoStat, error) {
--- a/src/common/resourceutil/cpu/cpu_test.go
+++ b/src/common/resourceutil/cpu/cpu_test.go
@@ -144,6 +144,36 @@ func removeFakeCPUMaxFreq() {
 	os.Remove("./fakecpumaxfreq")
 }
 
+func TestGetCPUFreqCpuInfo(t *testing.T) {
+	t.Run("Success", func(t *testing.T) {
+		originFileOpen = fileOpen
+		fileOpen = fakeFileOpenCPUInfo
+		defer func() {
+			fileOpen = originFileOpen
+			removeFakeCPUInfo()
+		}()
+
+		ret, err := getCPUFreqCpuInfo()
+		if err != nil {
+			t.Error("unexpected error")
+		} else if ret != 3300.0 {
+			t.Error("unexpected result")
+		}
+	})
+	t.Run("Error", func(t *testing.T) {
+		originFileOpen = fileOpen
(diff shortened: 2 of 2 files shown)


suresh-lc:
The value changes in cpuinfo. Hence this approach is not perfect and also it is not maximum cpu frequency

tdrozdovsky:
Yes you are right the meaning is changing. I wrote about it. There is no other way in virtual machines. Default is used '/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq' and if it is absent (when used in virtual machines), we can take inaccurate processor parameters. I think this is better than a system that does not work.

suresh-lc:
> @suresh-lc Yes you are right the meaning is changing. I wrote about it. There is no other way in virtual machines. Default is used '/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq' and if it is absent (when used in virtual machines), we can take inaccurate processor parameters. I think this is better than a system that does not work.
Home-Edge don't support kvm itself and the scenario was created since initial reporter didn't have the device. hence supporting with inaccurate by this approach would not be good one, as this is not normal flow(use case)

tdrozdovsky:
Yes, we never mentioned that we support Virtual Machines, but is it really bad if our system supports this? I think we should be more open to external developers.

suresh-lc:
I feel then we could reply on following param in the cpuinfo file
'model name	: Intel(R) Core(TM) i3-2120 CPU @ 3.30GHz'
Parse the above parameter and take the 3.30GHz and convert and use it. This is more in lines of "cpuinfo_max_freq" file

tdrozdovsky:
@suresh-lc  I considered this idea also, but it doesn't reflect the maximum value, it is the optimal processor frequency. See real example below:
```
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 58
model name	: Intel(R) Core(TM) i7-3635QM CPU @ 2.40GHz
stepping	: 9
microcode	: 0x21
cpu MHz		: 3273.342
cache size	: 6144 KB
physical id	: 0
```
So I didn't complicate parsing of /proc/cpuinfo.
If you have any idea please share it. 

I believe together we can make the Edge Orchestration better!

suresh-lc:
> @suresh-lc I considered this idea also, but it doesn't reflect the maximum value, it is the optimal processor frequency. See real example below:
> 
> ```
> processor	: 0
> vendor_id	: GenuineIntel
> cpu family	: 6
> model		: 58
> model name	: Intel(R) Core(TM) i7-3635QM CPU @ 2.40GHz
> stepping	: 9
> microcode	: 0x21
> cpu MHz		: 3273.342
> cache size	: 6144 KB
> physical id	: 0
> ```
> 
> So I didn't complicate parsing of /proc/cpuinfo.
> If you have any idea please share it.
> 
> I believe together we can make the Edge Orchestration better!

Following is the real example which we had crossed verified in many systems
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 42
model name	: Intel(R) Core(TM) i3-2120 CPU @ 3.30GHz
stepping	: 7
microcode	: 0x2f
cpu MHz		: 1926.375
cache size	: 3072 KB
physical id	: 0
siblings	: 4
core id		: 0
cpu cores	: 2
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 13
wp		: yes

tdrozdovsky:
@suresh-lc I agree with you, the value of CPU freq can be different (it depends from power mode cpu), especially in Virtual Machines. and now I am implementing the parser and tests of `model name : Intel(R) Core(TM) i3-2120 CPU @ 3.30GHz`. When everything is ready I will update PR.

P.S. I would like to note that by default the value of the maximum processor frequency will be read from the `/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq` and only in cases of failure from the mechanism proposed by me. Therefore, it will not affect PCs where the Edge Orchestration will run in the host Linux OS 

Thank you

tdrozdovsky:
@suresh-lc @MoonkiHong I've updated PR, please review.

PR #86, opened 2020-04-03
Add securemgr: container verification before executing (protection ag…

### LINKED ISSUE #85: Launch of malicious containers
**Describe the bug**

A critical system security vulnerability was discovered. In case of hacking of any device and the possibility of transferring REST API to the LF Edge network, attcker can get full access to the system of any device. as an example, the case of access to the file system is described below.

**To Reproduce**
1. Run Edge Orchestration container
```shell
$ docker run -it -d --privileged --network="host" --name edge-orchestration \
                -v /var/edge-orchestration/:/var/edge-orchestration/:rw \
                -v /var/run/docker.sock:/var/run/docker.sock:rw \
                -v /proc/:/process/:ro edge-orchestration:baobab
``` 

2. Send the next REST API request
```shell
curl -X POST "127.0.0.1:56001/api/v1/orchestration/services" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"ServiceName\": \"Simple-server\", \"ServiceInfo\": [{ \"ExecutionType\": \"container\", \"ExecCmd\": [ \"docker\", \"run\", \"-v\", \"/var/run:/var/run:rw\", \"-v\", \"/:/app\", \"-p\", \"80:8000\", \"-p\", \"emazzotta/simple-fileserver-docker\"]}], \"StatusCallbackURI\": \"http://localhost:8888/api/v1/services/notification\"}"
```
3. Start the browser and enter the ip address of the device on which the container was launched. 
After that, you will see the file system of the device under attack and get access to it

изображение

**Expected behavior**
Avoid launching unverified containers. Сreating a list of containers that can be run on the device.

**Test environment configuration:**
* Software version: Ubuntu 16.04
* Hardware: x86-64
* Toolchain: Docker and Go recommended version

tdrozdovsky:
…ainst malicious containers)

# Description

To protect _Edge Devices_ against running malicious containers to the `secure` mode, a white list (hashes of containers) was added that can be launched on the device. The virifier is the first functionality that the security module will provide.
in the future, verification should work by default, not only in `secure` mode

Fixes #85 

## Type of change

- [x] This change requires a documentation update (detail info will be done after the adoption of this concept)

# How Has This Been Tested?

1. Assembly `edge-orchestration` in `secure` mode by command:
```shell
$ ./build.sh secure
```
2. Run Edge Orchestration container
```shell
$ docker run -it -d --privileged --network="host" --name edge-orchestration \
                -v /var/edge-orchestration/:/var/edge-orchestration/:rw \
                -v /var/run/docker.sock:/var/run/docker.sock:rw \
                -v /proc/:/process/:ro edge-orchestration:baobab
``` 

3. Send the next REST API requests
  3.1 Successful hello-world container launch
```shell
curl -X POST "127.0.0.1:56001/api/v1/orchestration/services" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"ServiceName\": \"hello-world\", \"ServiceInfo\": [{ \"ExecutionType\": \"container\", \"ExecCmd\": [ \"docker\", \"run\", \"-v\", \"/var/run:/var/run:rw\", \"hello-world@sha256:fc6a51919cfeb2e6763f62b6d9e8815acbf7cd2e476ea353743570610737b752\"]}], \"StatusCallbackURI\": \"http://localhost:8888/api/v1/services/notification\"}"
```

  3.2 Failed emazzotta/simple-fileserver-docker container launch
```shell
curl -X POST "127.0.0.1:56001/api/v1/orchestration/services" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"ServiceName\": \"Simple-server\", \"ServiceInfo\": [{ \"ExecutionType\": \"container\", \"ExecCmd\": [ \"docker\", \"run\", \"-v\", \"/var/run:/var/run:rw\", \"-v\", \"/:/app\", \"-p\", \"80:8000\", \"-p\", \"emazzotta/simple-fileserver-docker\"]}], \"StatusCallbackURI\": \"http://localhost:8888/api/v1/services/notification\"}"
```

**Test Configuration**:
* Software version: Ubuntu 16.04, etc.)
* Hardware: x86-64
* Toolchain: Docker and Go recommended versions

# Checklist:

### DIFF
--- a/GoMain/src/main/main.go
+++ b/GoMain/src/main/main.go
@@ -29,6 +29,7 @@ import (
 	"controller/discoverymgr"
 	"controller/scoringmgr"
 	"controller/servicemgr"
+	"controller/securemgr"
 	executor "controller/servicemgr/executor/containerexecutor"
 
 	"orchestrationapi"
@@ -56,6 +57,7 @@ const (
 	configPath          = edgeDir + "/apps"
 	dbPath              = edgeDir + "/data/db"
 	certificateFilePath = edgeDir + "/data/cert"
+	containerWhiteListPath = edgeDir + "/data/cwl"
 
 	cipherKeyFilePath = edgeDir + "/user/orchestration_userID.txt"
 	deviceIDFilePath  = edgeDir + "/device/orchestration_deviceID.txt"
@@ -90,6 +92,10 @@ func orchestrationInit() error {
 	if buildTags == "secure" {
 		log.Println("Orchestration init with secure option")
 		isSecured = true
+	}
+
+	if isSecured {
+		securemgr.Init(containerWhiteListPath)
 	}
 
 	restIns := restclient.GetRestClient()
--- a/GoMain/src/main/main.go
+++ b/GoMain/src/main/main.go
@@ -153,6 +153,15 @@ func orchestrationInit() error {
 	}
 	ehandle := externalhandler.GetHandler()
 	ehandle.SetOrchestrationAPI(externalapi)
+	// external secure rest api
+	if isSecured {
+		securemgrexternalapi, err := securemgr.GetExternalAPI()
+		if err != nil {
+			log.Fatalf("[%s] Secure manager external api : %s", logPrefix, err.Error())
+		} else {
+			ehandle.SetSecuremgrAPI(securemgrexternalapi)
+		}
+	}
 	ehandle.SetCipher(dummy.GetCipher(cipherKeyFilePath))
 	restEdgeRouter.Add(ehandle)
 
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ RESTAPI
     ```
   ---
    If the `edge-orchestration` was assembled with `secure` option.
-   You need to add a digest (sha256) to the last parameter. `"hello-world@sha256:fc6a51919cfeb2e6763f62b6d9e8815acbf7cd2e476ea353743570610737b752\"`
+   You need to add a digest (sha256) to the last parameter. `"hello-world@sha256:fc6a51919cfeb2e6763f62b6d9e8815acbf7cd2e476ea353743570610737b752"`. More information about it you can find [here](doc/secure_manager.md).
    ```
    $ curl -X POST "IP:56001/api/v1/orchestration/services" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"ServiceName\": \"hello-world\", \"ServiceInfo\": [{ \"ExecutionType\": \"container\", \"ExecCmd\": [ \"docker\", \"run\", \"-v\", \"/var/run:/var/run:rw\", \"hello-world@sha256:fc6a51919cfeb2e6763f62b6d9e8815acbf7cd2e476ea353743570610737b752\"]}], \"StatusCallbackURI\": \"http://localhost:8888/api/v1/services/notification\"}"
    ```  
--- a/README.md
+++ b/README.md
@@ -208,8 +208,19 @@ RESTAPI
     ```
 - Curl Example:
     ```json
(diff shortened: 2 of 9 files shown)


suresh-lc:
@tdrozdovsky : Thanks for your interest to contribute towards strengthening the security of Home Edge. Went through the change you have submitted. Understand your approach is to validate the containers before executing them using the hash of containers. And current code has this basic validating feature. If any developers would like to use the edge orchestration with these changes, would be required to create the file with the details of container. Better approach is to give them a API to do this and this could be in your next item (road map). But it would be better to commit with this feature also. Home Edge is also expected to run on small device

tdrozdovsky:
@suresh-lc : "And current code has this basic validating feature." - please explain where this functionality is implemented and how to use it.

tdrozdovsky:
@suresh-lc : **"If any developers would like to use the edge orchestration with these changes, would be required to create the file with the details of container."** - Yes, the user will need to create a container white list file, for now, this functionality will work only in "secure" mode. And this means that only containers that are allowed will be executed on devices
P.S. Protection always introduces some inconvenience to the operation of systems. We have to accept it.

**Better approach is to give them a API to do this and this could be in your next item (road map).**
Yes, it is in the plans, but first you still need to implement key generation, generation and verification of digital signatures. We should not create a patch set of commits, but gradually make changes. These are best practices that show external developers that the project is live and is being implemented as an open source project. 

**"Home Edge is also expected to run on small device"** - Сould you describe more detail about these devices (how small they are: memory, performance)

Thank you for your comment

wansuyoo:
The edge-orchestration service has the function or ability to execute the container as requested.
In my opinion,
Preventing malicious container execution requests seems to be required user-side operational view.
Alternatively, I think it is a role in a container repository, such as DCT(Docker Content Trust).
https://docs.docker.com/engine/security/trust/content_trust/

tdrozdovsky:
@wansuyoo:The purpose of the contatiner white list is not only to start containers and check their integrity, but to restrict the launch of any containers except those that are allowed. Since there are cases when the original functionality of the container may be a vulnerability for our system.
And of course, in the future, we need to be able to filter the source of container loading and we should only consider trusted container sources such as DCT(Docker Content Trust).

Thank you for the comment. Your opinion is very important for us to make the system better.

suresh-lc:
> @suresh-lc : **"And current code has this basic validating feature."** - please explain where this functionality is implemented and how to use it?

I intend to mean that your submission has the basic validating feature of container.

tdrozdovsky:
I would like to clarify that my submission doesn't check the container but checks whether it is possible to run the given container on the _Edge Device_

tdrozdovsky:
**Additional info:** To check containers, a separate entity is needed that will check containers in docker container repository (check libraries and other software, for example, checking them in the CVE database). But this is not in scope of this project. 
In this project, we should only consider the filtering of docker container repository as advised @wansuyoo ("a container repository, such as DCT(Docker Content Trust)".

suresh-lc:
As wansuyoo had mentioned, in edge orchestration we are not trying to validate the container. But at the same time we can have in place a way to check the integrity of the container. Adding the container to list of containers and make it available to execute looks a rudimentary way, but as you say its starting point and we can go ahead with future enhancements.
Was looking into EdgeX Foundry and security module in the same(https://fuji-docs.edgexfoundry.org/Ch-Security.html). If you can look into their architecture, it looks something similar. They use key, secret store, API gateway and access credentials. So we might also look into such direction, might be mTower of yours could be used instead of the secret store. 
So how about we make a plan, high level requirements of how we are going to address the security issues in Home Edge. Then we can prioritize and implement them.

tdrozdovsky:
I completely agree with you.
1. I'm currently doing an external REST API for the user so that he can add container hash to the container white list.
2. Checking the integrity of the containers is a different task and it is now secondary (we have more important security challenges)
3. EdgeX Foundry security module - this architecture is typical, and unfortunately we do not have it at the moment. But we will implement it step by step.
4. As I mentioned at the workshop, we need a security storage, I examined **Vault** for a long time and I think that we will consider it for these purposes.
5. **mTover** cannot be considered for secure data storage as it is intended for MCU and our potential devices are ARM CPUs and sometimes x86 CPUs. I made prototypes on Raspberry and Hikey960 to provide H/W protection (based on TrustZone), but you noticed that these are not our goals (and in the current situation I agree with you and we will postpone this for later).
6. We completely lack authentication, and therefore the authorization of devices and users.

P.S. After implementing the REST API for the container white list, I will prepare a my proposal for protection Edge-Orchestration and we will be able to consider together with you what is our main priority.

PR #90, opened 2020-05-28
External rest api authentication

tdrozdovsky:
# Description

Adds authentication of external REST APIs using a JSON Web Token.
This is base version. 

## Type of change

# How Has This Been Tested?
1. Create a JWT, you can use the script by running it as shown below.
```shell
$ . tools/jwt_gen.sh
```
The generated token is exported to the shell environment variable: `EDGE_ORCHESTRATION_TOKEN`.

2. Use a JWT to execute any REST API
Example below:
```shell
$ curl -X POST "127.0.0.1:56001/api/v1/orchestration/securemgr" -H "accept: applicationnt-Type: application/json" -H "Authorization: $EDGE_ORCHESTRATION_TOKEN" -d "{ \"SecureMgr\": \"Verifier\", \"CmdType\": \"printAllHashCWL\", \"StatusCallbackURI\": \"http://localhost:8888/api/v1/services/notification\"}"
```
---
**Test Configuration**:
* Firmware version: (OS type and version, e.g., Ubuntu 14.04, etc.)
* Hardware: x86-64
* Toolchain: Docker and Go recommended version

# Checklist:


src/controller/securemgr/verifier/verifier.go
```diff
@@ -26,58 +26,50 @@ import (
...
-	SecureCmpName	string
+type ResponseVerifierConf struct {
+	Message       string
+	SecureCmpName string
 }
 
-// SecureMgrExternalAPI is the interface implemented by external REST API
-type SecureMgrExternalAPI interface {
-	RequestSecureMgr(containerInfo RequestSecureMgr) ResponseSecureMgr
+// VerifierConf is the interface implemented by external REST API
```
MoonkiHong, line 64: @tdrozdovsky @suresh-lc Any impact about this API change? It is just the call name change though, but it should be clear that : 1) previous SecureMgrExternalAPI itself is the API provided by SecureMgr, now 2) the proposed commit changes it to VerifierConf, which consists of one of the workflows (functions) in the territory of SecureMgr.

tdrozdovsky: The good and correct remark, but I think we should not worry about this since the external api has not changed. This is only the internal changes that relate to the handler (I decided to simplify our code and not introduce a separate api).
Regarding adding a token: this in any case will lead to the fact that any request will require adjustment. Unfortunately, protection always leads to difficulties in configuration and use. Especially if it has not been laid since the beginning of development. thanks for the comment

MoonkiHong: @tdrozdovsky Thanks for your comment. Overall I agree with your point. By the way, for things more clarified, is there any API document regarding this? It would be helpful for us to fully understand your opinion in the code level. Thank you in advance.

tdrozdovsky: Did you see the [doc/secure_manager.md](https://github.com/tdrozdovsky/edge-home-orchestration-go/blob/external-rest-api-authentication/doc/secure_manager.md) or you need more detailed information. Later I plan to add information to  `edge_orchestration_api.yaml`.

src/controller/securemgr/verifier/verifier.go
```diff
@@ -245,48 +237,48 @@ func printAllHashFromContainerWhiteList() {
 	}
 }
```
suresh-lc, line 239: Currently the JWT verification is being done for add, delete, deleteall and print hash functions. So the verification is basically with in the same system and when user tries to add the hash. How about adding similar security when System A tries to access System B for offloading.

src/controller/securemgr/verifier/verifier.go
MoonkiHong: @tdrozdovsky Yes. Better to present a bit more details like yaml files. But understood that everything comes at one time. Thank you.

src/controller/securemgr/verifier/verifier.go
tdrozdovsky: As I wrote to Suresh, basic authentication for the external api is implemeted. I will do internal api later (I added internal requests on the router to exceptions, it means that we don't check internal requests).

src/controller/securemgr/verifier/verifier.go
tdrozdovsky: @MoonkiHong @suresh-lc I corrected some inaccuracies in the description of the quick start (changes were made after the commit on February 18). And also **added REST API descriptions** for secure mode (**YAML format**)

mgjeong:
@suresh-lc 
Suresh, I'd like to finish the review if you agree on this change. 
@MoonkiHong , @tdrozdovsky 
Thank you for the constructive discussion.

MoonkiHong:
> Changes are fine

@suresh-lc Please approve this PR then! Thank you in advance!!

PR #94, opened 2020-06-20
Add initialization of verifier and authenticator in OrchestrationInit…

### LINKED ISSUE #93: OrchestrationInit() failed for java binding / c binding due to recent security changes
**Describe the bug**
Due to recent changes https://github.com/lf-edge/edge-home-orchestration-go/pull/90/files which mandate verifier to be enabled for all the platforms: Container, native and android. It makes orchestrationInit() API to be failed for interfaces (src/interfaces/javaapi and src/interfaces/capi) 

**To Reproduce**
1. Generate java object.
2. Include binding in application. 
3. Call orchestrationInit() API.  **(It is failing)**

**Expected behavior**
OrchestrationInit() should work for all the platform.

**Suggestion**
File Name: orchestration.go 
Line No: 162 (!o.isSetVerifierConf )

The above check should be done only for container platform. As this is common file for all the platform. 
Please check the common files for different platform and put necessary conditional checks.

tdrozdovsky:
… for native and android

# Description

Initially, code for working on other platforms was not added to the repository, since there is no description for these platforms. I added the initialization code in this PR, but it should be noted that I can not test it completely.

I’m happy to help you with a description or something else, just show me the starting point (for example, a link to the repository)

Fixes #93 

## Type of change

# How Has This Been Tested?

For testing, you need to add a description (or a link to the repository), an example of how to run, which compiler should be used for arm.

# Checklist:

### DIFF
--- a/src/interfaces/capi/main.go
+++ b/src/interfaces/capi/main.go
@@ -103,6 +103,8 @@ import (
 	configuremgr "controller/configuremgr/native"
 	"controller/discoverymgr"
 	scoringmgr "controller/scoringmgr"
+	"controller/securemgr/authenticator"
+	"controller/securemgr/verifier"
 	"controller/servicemgr"
 	"controller/servicemgr/executor/nativeexecutor"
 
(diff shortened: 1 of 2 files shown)


MoonkiHong:
@suresh-lc Please review this PR.

jay11ca39:
@tdrozdovsky ,
As per my understanding: 
How about enabling security only for container mode. Once it is stable we can plan about java and c binding ? 

Please provide your opinion ..

tdrozdovsky:
@jay11ca39 My opinion is as follows:
1. You need to add a description of how to test for other platforms
2. You can create branches from the master or switch to the tag where protection is not enabled (temporary solution, since protection needs to be added anyway)

P.S. It is very cool that our system can work on different platforms, let's add a description! This will help us to involve external developers

MoonkiHong:
@jay11ca39 @tdrozdovsky @suresh-lc Any further evaluation that we need?

MoonkiHong:
@suresh-lc PTAL..

tdrozdovsky:
Unfortunately, no one wrote how to test for platforms. I was forced to analyze the source code myself. I successfully built for the native platform, and also launched in protection mode. it works fine. It should be noted that under these platforms docker containers are not supposed to work, therefore the verifier does not take part for these platforms


P.S. For me it seems strange that here is a code that does not support the main feature of the orchestration system - containerization. And most importantly, there is no description how to use for other platforms.
1. I will add a description for the native platform soon.
2. For the android, I will do this later because it is more complicated and I need to analyze the code.

tdrozdovsky:
Colleagues, please tell me what needs to be done so that the code is accepted?

PR #108, opened 2020-07-14
Add signalmgr to manage the edge orchestrator by signal

### LINKED ISSUE #107: Use signals instead of timer
How about changing [time.sleep functions](https://github.com/lf-edge/edge-home-orchestration-go/blob/master/GoMain/src/main/main.go#L169) to the code controlled by signal such as SIGINT, SIGTERM?
I think it could expand other scenarios such as 
 - A node sends it's termination request to other neighbors if the node receives SIGTERM.

t25kim:
Signed-off-by: t25kim <t25.kim@samsung.com>

# Description

As [@MoonkiHong's advise](https://github.com/lf-edge/edge-home-orchestration-go/issues/107#issuecomment-657982824), I've made the signal manager at the common folder.
This PR includes initial step of sigmgr code and the scenario that the edge orchestrator will be terminated if sigmgr detects SIGINT, SIGKILL and SIGTERM.
Fixes #107 

## Type of change

# How Has This Been Tested?

This PR has been tested as below.
1. Run the edge orchestrator
2-1. Produce SIGTERM ( $ docker stop edge-orchestration )
2-2. Produce SIGINT ( input "Ctrl + c" at the edge-orchestration container )

**Test Configuration**:
* Firmware version: Ubuntu 16.04
* Hardware: x86-64

# Checklist:

- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/GoMain/src/main/main.go
+++ b/GoMain/src/main/main.go
@@ -21,9 +21,9 @@ package main
 import (
 	"errors"
 	"log"
-	"time"
 
 	"common/logmgr"
+	"common/sigmgr"
 
 	configuremgr "controller/configuremgr/container"
 	"controller/discoverymgr"
@@ -74,10 +74,7 @@ func main() {
 	if err := orchestrationInit(); err != nil {
 		log.Fatalf("[%s] Orchestaration initalize fail : %s", logPrefix, err.Error())
 	}
-
-	for {
-		time.Sleep(1000)
-	}
+	sigmgr.Watch()
 }
 
 // orchestrationInit runs orchestration service and discovers other orchestration services in other devices
@@ -163,10 +160,6 @@ func orchestrationInit() error {
 	restEdgeRouter.Start()
 
 	log.Println(logPrefix, "orchestration init done")
-
-	// TODO remove this line
-	// this line is for wait to initialize the mDNS server.
-	time.Sleep(time.Second * 2)
 
 	return nil
 }
--- a/GoMain/run.sh
+++ b/GoMain/run.sh
@@ -1,2 +1,10 @@
 #!/bin/bash
-/edge-orchestration/edge-orchestration
+/edge-orchestration/edge-orchestration &
+
+edge_pid=$!
+# Trap to forward SIGTERM to the child edge container
+trap 'kill -TERM $edge_pid' TERM
+wait $edge_pid
+# Restore SIGTERM to its default behavior
+trap - TERM
+wait $edge_pid
--- a/build.sh
+++ b/build.sh
@@ -16,6 +16,7 @@ PKG_LIST=(
         "common/resourceutil/cpu"
         "common/resourceutil/native"
         "common/resourceutil/types/servicemgrtypes"
+        "common/sigmgr"
         "common/types/configuremgrtypes"
         "common/types/servicemgrtypes"
         "controller/configuremgr"
--- a/glide.yaml
+++ b/glide.yaml
@@ -2,6 +2,10 @@ package: .
 excludeDirs:
   - CMain
 ignore:
+  - common/commandvalidator
+  - common/commandvalidator/blacklist
+  - common/commandvalidator/commands
+  - common/commandvalidator/injectionchecker
   - common/errormsg
   - common/errors
   - common/logmgr
@@ -13,12 +17,9 @@ ignore:
   - common/resourceutil/cpu
   - common/resourceutil/native
   - common/resourceutil/types/servicemgrtypes
+  - common/sigmgr
   - common/types/configuremgrtypes
   - common/types/servicemgrtypes
-  - common/commandvalidator
-  - common/commandvalidator/blacklist
-  - common/commandvalidator/commands
-  - common/commandvalidator/injectionchecker
   - controller/configuremgr
   - controller/configuremgr/container
   - controller/configuremgr/native
--- a/src/common/sigmgr/sigmgr.go
+++ b/src/common/sigmgr/sigmgr.go
@@ -1,0 +1,37 @@
+/*******************************************************************************
+ * Copyright 2020 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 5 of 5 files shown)


MoonkiHong:
@suresh-lc PTAL.

GoMain/src/main/main.go
```diff
@@ -21,9 +21,9 @@ package main
 import (
 	"errors"
 	"log"
-	"time"
 
 	"common/logmgr"
+	"common/sigmgr"
```
MoonkiHong, line 26: @suresh-lc We all recognize that `configuremgr` provides `watcher` to notify any state change of services on top of `edge-home-orchestration-go`. But this is for service notification. This PR suggests a separate manager called `sigmgr` to deal with Termination scenario of the edge-home-orchestration with respect to the newly suggested `SIGTERM`. What do you think of this approach from architecture perspective. IMHO, I think that this approach is quite valid.

GoMain/src/main/main.go
```diff
@@ -74,10 +74,7 @@ func main() {
 	if err := orchestrationInit(); err != nil {
 		log.Fatalf("[%s] Orchestaration initalize fail : %s", logPrefix, err.Error())
 	}
-
-	for {
-		time.Sleep(1000)
-	}
+	sigmgr.Watch()
```
MoonkiHong, line 77: @suresh-lc I think that this is quite valid because the previous implementation is hard-coded with the `Sleep` func with a specific value.

GoMain/src/main/main.go
```diff
@@ -164,9 +161,5 @@ func orchestrationInit() error {
 
 	log.Println(logPrefix, "orchestration init done")
 
-	// TODO remove this line
```
MoonkiHong, line 167: @suresh-lc LGTM. What do you think of deprecating the hard-coded `Sleep` operation?

GoMain/src/main/main.go
suresh-lc: @t25kim : So the changes suggested here is for abnormal termination of edge-orchestration-go itself. From testing point, i understand it has edge-orchestration is started as docker container and stopped abruptly. So the abnormal stop gets handled. In Such case is the port also gets closed, can you check that once if not checked.

tdrozdovsky:
I like the idea of adding a signal handler to the code and the direction of the signals to `edge-orchestration` child process  from `sh`. As I mentioned earlier, signal processing should exclude errors that may occur while working with the database and others.

GoMain/src/main/main.go
t25kim: @suresh-lc Thank you for the comment. This sigmgr could apply both of normal and abnormal termination scenarios.
For the normal example, if a user, who wants to terminate a edge container, run "docker stop edge-orchestration" and docker will produce SIGTERM and SIGKILL to stop it gracefully. [[REF]](https://docs.docker.com/engine/reference/commandline/stop/)

In this PR, I have checked the edge-orchestration goes to exit status when SIGTERM, SIGINT signals reach to the edge-orchestration process.

- Case 1. $ docker stop edge-orchestration
```
t25kim@t25kim:~$ docker ps
CONTAINER ID        IMAGE                       COMMAND             CREATED             STATUS              PORTS               NAMES
174666e6fe34        edge-orchestration:baobab   "sh run.sh"         5 seconds ago       Up 4 seconds                            edge-orchestration
t25kim@t25kim:~$ docker stop edge-orchestration
edge-orchestration
t25kim@t25kim:~$ docker ps -a
CONTAINER ID        IMAGE                       COMMAND                  CREATED             STATUS                      PORTS               NAMES
174666e6fe34        edge-orchestration:baobab   "sh run.sh"              39 seconds ago      Exited (0) 8 seconds ago                        edge-orchestration
```
- Case 2.  "Ctrl + c" at the edge-orchestration process in the container
```
t25kim@t25kim:~$ docker ps
CONTAINER ID        IMAGE                       COMMAND             CREATED             STATUS              PORTS               NAMES
552e88e345a7        edge-orchestration:baobab   "sh run.sh"         13 seconds ago      Up 13 seconds                           edge-orchestration
t25kim@t25kim:~$ docker ps -a
CONTAINER ID        IMAGE                       COMMAND                  CREATED             STATUS                       PORTS               NAMES
552e88e345a7        edge-orchestration:baobab   "sh run.sh"              31 seconds ago      Exited (130) 5 seconds ago                       edge-orchestration
```

- Case 3. SIGTERM to the edge-orchestration process in the container
```
t25kim@t25kim:~$ docker ps
CONTAINER ID        IMAGE                       COMMAND             CREATED             STATUS              PORTS               NAMES
ae2c43b75afe        edge-orchestration:baobab   "sh run.sh"         6 seconds ago       Up 5 seconds                            edge-orchestration
t25kim@t25kim:~$ docker exec -it edge-orchestration /bin/bash
root@t25kim:/edge-orchestration# ps -ef
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 02:08 pts/0    00:00:00 sh run.sh
root         6     1  0 02:08 pts/0    00:00:00 /edge-orchestration/edge-orchestration
root        18     0  0 02:09 pts/1    00:00:00 /bin/bash
root        27    18  0 02:09 pts/1    00:00:00 ps -ef
root@t25kim:/edge-orchestration# kill -15 6
root@t25kim:/edge-orchestration# 
t25kim@t25kim:~$ docker ps -a
CONTAINER ID        IMAGE                       COMMAND                  CREATED             STATUS                      PORTS               NAMES
ae2c43b75afe        edge-orchestration:baobab   "sh run.sh"              37 seconds ago      Exited (0) 15 seconds ago                       edge-orchestration
```

t25kim:
> I like the idea of adding a signal handler to the code and the direction of the signals to `edge-orchestration` child process from `sh`. As I mentioned earlier, signal processing should exclude errors that may occur while working with the database and others.

@tdrozdovsky I totally agree with you. By signal processing, edge-orchestration must exclude errors and inform neighbors of what errors happened.

GoMain/src/main/main.go
MoonkiHong: @suresh-lc LGTM to accept this PR.

PR #134, opened 2020-09-21
Change file permissions from 755 to 644

t25kim:

# Description

This PR changes file permissions from 755 to 644 since the files (such as go, md and etc) should not have execute permission.

## Type of change

- [x] Bug fix 

# How Has This Been Tested?
Run the build script 
```
$ ./build.sh
```

**Test Configuration**:
* Firmware version: Ubuntu 18.04
* Hardware: x86-64

# Checklist:


MoonkiHong (commented):
LGTM, since it does not have any change but granting the file permissions from 100755 to 100644.

MoonkiHong (approved):
LGTM.

MoonkiHong:
@suresh-lc @Karthikeyan-Samsung PTAL.

suresh-lc:
While i was trying to check on this found this info :Permissions of 644 mean that the owner of the file has read and write access, while the group members and other users on the system only have read access. For executable files, the equivalent settings would be 700 and 755 which correspond to 600 and 644 except with execution permission." 
Understand that in this PR is removing execute permission.  Any motive behind it. 
Will further check if we need for the change.

suresh-lc:
If the intention is to remove the execute permission to users for files, instead of changing for individual files how about changing for the directory alone. home directory drwx------ (700). If we do that, we don't need to protect files individually. This won't break anything other than the ability of other people to read the file.

t25kim:
@suresh-lc Thank you for the review!
This PR is not intended to remove the file permissions from other users.
I thought all the files should have correct permissions including the owner because of the [vulnerability issue](https://cwe.mitre.org/data/definitions/732.html).

PR #157, opened 2020-10-29
Fix copyright date

tdrozdovsky:

# Description

The correction concerns the copyright date. Since according to the rules, after changing the file, the year of change is added and not corrected. for example the file was created in 2016, in 2019 and in 2020 was edited then the entry should be 2016, 2019-2020.
This must be strictly followed as this can lead to ownership disputes.

P.S. If you know more files where this was fixed from 2019 to 2020, please make the edits.

## Type of change

# Checklist:

### DIFF
--- a/src/interfaces/capi/main.go
+++ b/src/interfaces/capi/main.go
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright 2020 Samsung Electronics All Rights Reserved.
+ * Copyright 2019-2020 Samsung Electronics All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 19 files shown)


MoonkiHong:
@tdrozdovsky Thank you. Will back to you soon after our `Coconut` release.

src/interfaces/capi/main.go
```diff
@@ -19,7 +19,7 @@
 package main
 
 ///*******************************************************************************
-// * Copyright 2020 Samsung Electronics All Rights Reserved.
+// * Copyright 2019-2020 Samsung Electronics All Rights Reserved.
```
MoonkiHong, line 22: @tdrozdovsky What about removing this statement due to the duplication of the state at the header?

MoonkiHong (commented):
@tdrozdovsky Please see my comment (for a suggestion to revise)

src/interfaces/capi/main.go
tdrozdovsky: I believe that this is a duplication, but on the other hand it is an attachment of С file that has its own license.

I didn't create this file and I don't know what the author meant. So I left the code unchanged.

MoonkiHong: @tdrozdovsky Uhm... This is not a matter of who creates the file, but the original proposal of this PR is to rectify the copyright statement of each updated / created file in our project and I do not believe that the current 19 files are not created from you, but appreciatively, your PR suggests to rectify the statement. 

All the copyright statements of the files in our project are described at the top most of each file and there is no exception.

MoonkiHong (changes requested):
PTAL.

src/interfaces/capi/main.go
tdrozdovsky: @MoonkiHong I've removed the Copyright header.

MoonkiHong (approved):
LGTM.

suresh-lc (approved):
Thanks for incorporating in all 19 files

Issue #168, opened 2020-11-06
Configure secure and mnedc mode while running the image


t25kim:
**Is your feature request related to a problem? Please describe.**
Currently, the edge-orchestration sets the secure and mnedc options while building the project.

**Describe the solution you'd like**
Read the environment values if the edge-orchestration runs with SECURE and MNEDC options like below.
```
docker run -it -d --privileged ... -e SECURE=true -e MNEDC=server edge-orchestration:coconut
```

MoonkiHong:
@t25kim Thank you for the good suggestion. It seems good to me to estimate further for this proposal.

PR #170, opened 2020-11-09
Update the doc related to environment variables

t25kim:

# Description

Update the doc related to environment variables.

Related Issue #168 

## Type of change

- [x] Documentation update

# How Has This Been Tested?

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code

### DIFF
--- a/doc/platforms/x86_64_linux/x86_64_linux.md
+++ b/doc/platforms/x86_64_linux/x86_64_linux.md
@@ -143,6 +143,22 @@ Note that you can visit [Swagger Editor](https://editor.swagger.io/) to graphica
 ```shell
 $ docker run -it -d --privileged --network="host" --name edge-orchestration -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro edge-orchestration:coconut
 ```
+- Environment Variables
+
+  You can adjust the configuration of the edge-orchestration container by passing the environment variables below on the docker run command.
+  - SECURE
+
+    [Secure](../../secure_manager.md) mode can be enabled by setting SECURE to `true`.
+    ```shell
+    $ docker run -it -d --privileged --network="host" --name edge-orchestration -e SECURE=true -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro edge-orchestration:coconut
+    ```
+  - MNEDC
+
+    [MNEDC](../../mnedc.md) mode can be enabled by setting MNEDC to `server` or `client`.
+    ```shell
+    $ docker run -it -d --privileged --network="host" --name edge-orchestration -e MNEDC=server -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro edge-orchestration:coconut
+    ```
+
 - Result
 
 ```shell


MoonkiHong (approved):
Document updates for applying the build options. LGTM.

tdrozdovsky:
The ability to configure the edge-orchestration from the command line makes it more flexible.
But enabling or disabling protection (without secure) makes the edge-orchestration more vulnerable, as it enables the user (hacker) to disable protection. As I said earlier, the operation of the edge-orchestration without protection should be temporary, only at the development stage (for simplicity), by default, the protection must be enabled.

You need to think again whether you need to use the ability to start the edge-orchestratin without protection. If you have weighty arguments, please write to me.

Temporarily this solution may be, but I think in the future it will be necessary to revert to the previous version or cancel the ability to build without secure.

Please note that my assumption only applies to protected mode.

MoonkiHong:
> @MoonkiHong @t25kim The ability to configure the edge-orchestration from the command line makes it more flexible.
> But enabling or disabling protection (without secure) makes the edge-orchestration more vulnerable, as it allows the user (hacker) to disable protection. As I said earlier, the operation of the edge-orchestration without protection should be temporary, only at the development stage (for simplicity), by default, the protection must be enabled.
> 
> You need to think again whether you need to use the ability to start the edge-orchestration without protection. If you have weighty arguments, please write to me.
> 
> Temporarily this solution may be, but I think in the future it will be necessary to revert to the previous version or cancel the ability to build without secure.
> 
> Please note that my assumption only applies to protected mode.
> 
> it is very important for me to hear your opinion on this request

@tdrozdovsky Understood that the `secure` mode is by default, but we are a long way to go. Anyway, the initial proposal to setup the `secure` mode is a separate option to build for it. As you suggested, this is a temporary version, but we need to clearly indicate that our **regular** release should be by default with the `secure` mode. Thought?

@tdrozdovsky But still even with `secure` mode by default, as you see our recent security analysis provided by LGTM, there are 17 alerts/issues that we need to fix for the secure operation. (Of course, most of them are for the debugging mode, but this is anyhow related to the security be design.) Hope to get your feedback further. (aside from evaluating this PR)

tdrozdovsky:
@MoonkiHong Yes, I completely agree with you, I just want us to remember this and not accept that this is a permanent decision. As I said, I like that we have flexibility in configuration.

t25kim:
> @MoonkiHong @t25kim The ability to configure the edge-orchestration from the command line makes it more flexible.
> But enabling or disabling protection (without secure) makes the edge-orchestration more vulnerable, as it allows the user (hacker) to disable protection. As I said earlier, the operation of the edge-orchestration without protection should be temporary, only at the development stage (for simplicity), by default, the protection must be enabled.
> 
> You need to think again whether you need to use the ability to start the edge-orchestration without protection. If you have weighty arguments, please write to me.
> 
> Temporarily this solution may be, but I think in the future it will be necessary to revert to the previous version or cancel the ability to build without secure.
> 
> Please note that my assumption only applies to protected mode.
> 
> it is very important for me to hear your opinion on this request

@tdrozdovsky Thank you very much for your valuable idea.
Please don't get this issue wrong. This item is intended to support the user convenience since building several edge-orchestration images is way too inefficient. 
It would be good to write down to use `secure` mode for commercialization in the README file.

suresh-lc (approved):
LGTM. By default Secure would be enabled as the conversations.

MoonkiHong:
> LGTM. By default Secure would be enabled as the conversations.

@suresh-lc Just for a clarification, the current source code is not built as a secure mode by default, it is enabled by `secure` option. This PR is for a temporary approach before we get full features described by Wiki (all 4~5 modules). What we have discussed so far is: we would like to enable the `secure` mode by default in the future  **regular** release for those who would like to commercialize their product with Home Edge.

PR #189, opened 2020-12-01
Fix accessible confidential information that is recorded in unencrypt…

### LINKED ISSUE #141: [WIP] Encryption of SecureMgr
_Originally posted by @tdrozdovsky in https://github.com/lf-edge/edge-home-orchestration-go/issues/131#issuecomment-691930137_

> 
> 
> I personally think that sensitive data is returned by an access to `passPhraseJWTPath`. What do you think? @tdrozdovsky
> 
> https://github.com/lf-edge/edge-home-orchestration-go/blob/22ce49b4b76ae78dfb75e14ef64399e5805c7d02/src/controller/securemgr/authenticator/authenticator.go#L70
> 
> This might be the same potential security risk as follow.
> 
> https://github.com/lf-edge/edge-home-orchestration-go/blob/22ce49b4b76ae78dfb75e14ef64399e5805c7d02/src/controller/securemgr/authenticator/authenticator.go#L84

Good point, I know and remember this security issue.

These only informs about a failed attempt to create the `passPhraseJWTFilePath` file.
But of course, storing such information in files (passPhrase, edge-orchestration.key, etc) is a security risk.
I think in the future this should be solved with secure storage or with access control system such as: SeLinux, SMACK, etc.

Thank you for reminder

tdrozdovsky:
…ed form

# Description

File names were removed from logging in accordance with the Common Weakness Enumeration: CWE-312, CWE-315. In this case, there were no security threats, but automatic vulnerability analysis systems recommend doing this.

Fixes #141

## Type of change

# How Has This Been Tested?

Run LGTM checker

**Test Configuration**:
* Firmware version: Ubuntu 16.04, etc.)
* Hardware: x86-64
* Toolchain: Docker and Go recommended versions

# Checklist:

### DIFF
--- a/src/controller/securemgr/authenticator/authenticator.go
+++ b/src/controller/securemgr/authenticator/authenticator.go
@@ -68,7 +68,7 @@ func Init(passPhraseJWTPath string) {
 	if _, err := os.Stat(passPhraseJWTPath); err != nil {
 		err := os.MkdirAll(passPhraseJWTPath, os.ModePerm)
 		if err != nil {
-			log.Panicf("Failed to create passPhraseJWTPath %s: %s\n", passPhraseJWTPath, err)
+			log.Panicf("%s Failed to create passPhraseJWTPath: %s\n", logPrefix, err)
 			return
 		}
 	}
@@ -82,7 +82,7 @@ func Init(passPhraseJWTPath string) {
 		passphrase = []byte(randString(16))
 		err = ioutil.WriteFile(passPhraseJWTFilePath, passphrase, 0666)
 		if err != nil {
-			log.Println(logPrefix, "cannot create "+passPhraseJWTFilePath+": ", err)
+			log.Println(logPrefix, "Cannot create passPhraseJWT.txt:", err)
 		}
 	}
 


t25kim (approved):
LGTM!

MoonkiHong (approved):
This is great! LGTM.

suresh-lc (approved):
LGTM

src/controller/securemgr/authenticator/authenticator.go
```diff
@@ -82,7 +82,7 @@ func Init(passPhraseJWTPath string) {
 		passphrase = []byte(randString(16))
 		err = ioutil.WriteFile(passPhraseJWTFilePath, passphrase, 0666)
 		if err != nil {
-			log.Println(logPrefix, "cannot create "+passPhraseJWTFilePath+": ", err)
+			log.Println(logPrefix, "Cannot create passPhraseJWT.txt:", err)
```
mgjeong, line 85: Looks good, but ```'passPhraseJWT.txt'``` also looks like a filename( information).
What about you use just a simple text like ```'passphrase file'```?

mgjeong (approved):
Good job.

src/controller/securemgr/authenticator/authenticator.go
tdrozdovsky: @mgjeong Thank you for your good question.
As I wrote earlier, this file should contain a phrase for generating a SHA-based JWT (we recommend using an RSA-based JWT). And this is described in the documentation, hence the attacker knows this.
If you look at the threat tree, then you should **divide the type of attack into remote and with direct access to the device**. At the current moment, the use of the token allows us to **protect against a remote attack**, but not against an attack when there is access to the device.
Of course, our goal is to provide maximum protection, and we will do this step by step.
In the future, I suppose we will store sensitive information in a secure storage.

mgjeong: Okay. I got it.
Thanks.

PR #250, opened 2021-01-27
Fix important warnings that were found by the linter

tdrozdovsky:

# Description

Fix important warnings that were found by the linter.

## Type of change

# How Has This Been Tested?

```
  - [x] Unittest
  - [x] Execution of Container
```

**Test Configuration**:
* Firmware version: Ubuntu 16.04
* Hardware: x86-64

# Checklist:

### DIFF
--- a/src/controller/securemgr/authorizer/authorizer.go
+++ b/src/controller/securemgr/authorizer/authorizer.go
@@ -14,14 +14,17 @@
 * limitations under the License.
 *
 *******************************************************************************/
+
+// Package authorizer provides Role Based Access Control (RBAC)
 package authorizer
 
 import (
 	"errors"
 	"io/ioutil"
-	"github.com/lf-edge/edge-home-orchestration-go/src/common/logmgr"
 	"net/http"
 	"os"
+
+	"github.com/lf-edge/edge-home-orchestration-go/src/common/logmgr"
 
 	"github.com/casbin/casbin"
 )
@@ -31,10 +34,10 @@ type AuthorizerImpl struct{}
 
 const (
 	rbacPolicyFileName = "policy.csv"
-	policy_template    = "p, admin, /*, *\n" +
+	policyTemplate     = "p, admin, /*, *\n" +
 		"p, member, /api/v1/orchestration/services, *\n"
 	rbacAuthModelFileName = "auth_model.conf"
-	auth_model_template   = "[request_definition]\n" +
+	authModelTemplate     = "[request_definition]\n" +
 		"r = sub, obj, act\n\n" +
 		"[policy_definition]\n" +
 		"p = sub, obj, act\n\n" +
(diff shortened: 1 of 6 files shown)


src/controller/securemgr/authorizer/authorizer.go
```diff
@@ -88,15 +91,15 @@ func Init(rbacRulePath string) {
 	}
 	rbacPolicyFilePath = rbacRulePath + "/" + rbacPolicyFileName
 	if _, err := os.Stat(rbacPolicyFilePath); err != nil {
-		err = ioutil.WriteFile(rbacPolicyFilePath, []byte(policy_template), 0666)
+		err = ioutil.WriteFile(rbacPolicyFilePath, []byte(policyTemplate), 0666)
```
t25kim, line 94: Sorry for other topic...
How about changing the file permission from 0666 to 0664?
I think granting write permission to any other users is unnecessary.

tdrozdovsky: > 
> 
> Sorry for other topic...
> How about changing the file permission from 0666 to 0664?
> I think granting write permission to any other users is unnecessary.

@t25kim Done! Thank you for your comment!

t25kim (approved):
LGTM. Thanks!

MoonkiHong (approved):
LGTM.

src/controller/securemgr/verifier/verifier.go
```diff
@@ -141,10 +144,9 @@ func (VerifierImpl) ContainerIsInWhiteList(containerName string) error {
 	if containerHashIsInWhiteList(containerName[index:]) {
 		log.Printf("%s container's hash: %s is in container white list\n", logPrefix, containerName[index:])
 		return nil
```
suresh-lc, line 146: In case of error, we return new error object, whereas in case of validated container we are returning nil. Some other logical return would be better i guess.

tdrozdovsky: Thanks for the comment! To return a different value, we need to add another return parameter. Could you give an example that you would like to see?

suresh-lc (approved):
LGTM

src/controller/securemgr/verifier/verifier.go
t25kim: As @suresh-lc's comment, it would be good to change some code. 
How about modifying as below?

1. func [containerHashIsInWhiteList](https://github.com/lf-edge/edge-home-orchestration-go/blob/beba02276bd9a563842b6b17d82c1ee84be29615/src/controller/securemgr/verifier/verifier.go#L104) :  return type from `bool`  to `error`
2. Move error logic here
```suggestion
		return errors.New("container's hash is not in container white list")
```
3. Move `return nil` to the last

suresh-lc: Got the point as suggested by @t25kim , but still we would return the nil. Since the function signature is to return error, guess returning nil would only suit it. Else we need to change the function signature might be something like

func vals() (int, int) {
    return 3, 7
}

and use whatever is needed like
 _, c := vals()

tdrozdovsky: As always, the Truth is in the middle.
1) I agree with the @t25kim and this simplifies the code at the moment
2) but for the future, I planned that the description of containers will contain additional information fields and here the @suresh-lc opinion is also correct.

Summary:
1) the code is working correctly now
2) I took note of your opinions as they are correct. Thank you!
3) In the near future I will think about whether we will limit more the container's policy. And we will discuss it again

t25kim (approved):
LGTM. Thanks!

PR #295, opened 2021-04-19
Add guide "how to add new platform"

### LINKED ISSUE #127: [Consideration] Policy for platform support
Currently, we have supported 1) x86-64 Linux, 2) Rpi 2, and 3) Hikey960 if we adopt the recent PR #125 . It is now a good timing to consider how we setup a policy to support those existing platform and additional ones in the future release including Coconut.

Any thoughts on this?

tdrozdovsky:
Signed-off-by: Taras Drozdovskyi <t.drozdovsky@gmail.com>

This PR provides a guide for adding a new platform. 

Fixes #127 

## Type of change

- [x] Documentation update

# Checklist:


docs/how-to-add-new-platform.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Contents
+1. [Introduction](#1-introduction)
+2. [Add a new platform](#2-add-a-new-platform)  
+    2.1 [Additional code structure](#21-additional-code-structure)  
+    2.2 [Update README.md](#22-update-readmemd)  
+3. [Code inclusion to Edge-Orchestration upstream](#3-code-inclusion-to-edge-orchestration-upstream)
+
+## 1. Introduction
+This document provides a description of how to add a new platform to the Edge-Orchestration project.
```
t25kim, line 11: How about using `Home Edge` instead of Edge-Orchestration?
As I know, edge-orchestration is an executable file and is not a project name.

t25kim (commented):
Thanks for the contribution and please take a look at my comment.

docs/how-to-add-new-platform.md
tdrozdovsky: Until that time, I don't understand exactly what the name of our project is. )

At the moment I am thinking of keeping the Edge-Orchestration name, since it is used in many places. But in general, I agree with you that the name needs to be clarified and changed in all places

tdrozdovsky: изображение

We have this name indicated here. )

t25kim: That's a good point.
How about calling LF Home Edge as an umbrella project? 
Then https://github.com/lf-edge/edge-home-orchestration-go could be called the Edge Orchestration project under Home Edge.

t25kim (approved):
LGTM!

docs/how-to-add-new-platform.md
tdrozdovsky: I think this is a difficult issue (choice of name) and we need to discuss it. Considering that the name is used in many refs. 

P.S. The umbrella project already exists. (

MoonkiHong: When it comes to this PR, it looks good to me to go as the current name of "edge-orchestration". When it comes to the umbrella, we can discuss more once there is a separate repo for HomeEdge aside from edge-home-orchestration-go. We thought that DataStorage and MNEDC could be from separate repos, but now it is integrated to the existing edge-home-orchestration-go. So we should carefully analyze if we keep the current repo architecture or create a separate repo for the new distinctive feature like CloudSync or so on. 

In short, as far as this PR is concerned, it looks good to me.

MoonkiHong (approved):
+1

PR #301, opened 2021-05-12
[DataStorage] API for String , Images (png, jpeg)

### LINKED ISSUE #286: [Data Storage] Provide an API in Home Edge to store device generated data
Existing Problem:
Currently if any developer (using Rpi/Sensor) want to store device generated data using Home Edge, it is not possible to do. The developer has to again use EdgeX API/SDK to store the data. This makes the developer to know the working of both Home Edge and EdgeX. Home Edge should provide seamless discovery, orchestration and data sharing mechanism. For developer Home Edge should be the only point of consideration and need not bother on other solutions used internally. 

Solution:
1. Home Edge can provide an API ( POST) which can be used by developers to store the device data. The API should provide methods to store int, float, string, binary(image) data. This API can internally leverage on EdgeX to store the data.
2. We should also have an APi to get the stored values of a particular device .

sun-sharma:
- Added Config to Store String type data.
- Modified the code to be able to store image data (jpeg/png)as Base64 strings.

Signed-off-by: Sunchit Sharma <sun.sharma@samsung.com>

# Description

-Use POST api like "IP:49986/api/v1/resource/datastorage/{string/png/jpeg}" to upload data
### String Example
curl --location --request POST 'http://localhost:49986/api/v1/resource/datastorage/string' \
--data-raw 'Hello World'

### PNG Example
curl --location --request POST 'http://localhost:49986/api/v1/resource/datastorage/png' \
--header 'Content-Type: image/png' \
--data-binary '@image.png'

### JPEG Example
curl --location --request POST 'http://localhost:49986/api/v1/resource/datastorage/jpeg' \
--header 'Content-Type: image/jpeg' \
--data-binary '@image.jpg'

Fixes #286 

## Type of change

# How Has This Been Tested?

1. Run edgex containers by docker-compose like below.
```
docker-compose -f deployments/datastorage/docker-compose.yml up -d
```
2. Copy DataStorage configuration files(configuration.toml and datastorage-device.yaml) to /var/edge-orchestration/datastorage/
3. Run edge-orchestration
4. Send some requests to upload data like below

```
curl --location --request POST 'http://localhost:49986/api/v1/resource/datastorage/string' \
--data-raw 'Samsung'
```

```
curl --location --request POST 'http://localhost:49986/api/v1/resource/datastorage/png' \
--header 'Content-Type: image/png' \
--data-binary '@image.png'
```

5. Check the results from edgex-core-data via a web browser like below :
 http://localhost:48080/api/v1/event/device/datastorage/2

Get

As we can see, strings are normally stored and images can be converted back using a Base64 decoder or online tools. This was done to verify that proper retrieval is possible post storage.

**Test Configuration**:
* Hardware: x86-64
* Edge Orchestration Release: v1.0.0

# Checklist:

- [X] My changes generate no new warnings

### DIFF
--- a/internal/controller/storagemgr/storagedriver/storagehandler.go
+++ b/internal/controller/storagemgr/storagedriver/storagehandler.go
@@ -19,6 +19,7 @@ package storagedriver
 
 import (
 	"context"
+	b64 "encoding/base64"
 	"fmt"
 	"io/ioutil"
 	"math"
@@ -196,6 +197,12 @@ func deviceHandler(writer http.ResponseWriter, request *http.Request) {
 	handler.processAsyncRequest(writer, request)
 }
 
+func convertToBase64(val []byte) string {
+	// Convert to Base 64
+	Base64Val := b64.StdEncoding.EncodeToString(val)
+	return Base64Val
+}
+
 func (handler StorageHandler) newCommandValue(resourceName string, reading interface{}, valueType models.ValueType) (*models.CommandValue, error) {
 	var result = &models.CommandValue{}
 	var errn error
@@ -214,7 +221,13 @@ func (handler StorageHandler) newCommandValue(resourceName string, reading inter
 		if !ok {
 			return nil, fmt.Errorf(castError, resourceName, "not []byte")
 		}
-		result, errn = models.NewCommandValue(resourceName, timestamp, val, valueType)
+		if resourceName == "jpeg" || resourceName == "png" {
+			Base64Val := convertToBase64(val)
+			valueType = models.String
+			result, errn = models.NewCommandValue(resourceName, timestamp, Base64Val, valueType)
+		}else {
+			result, errn = models.NewCommandValue(resourceName, timestamp, val, valueType)
+		}
 
 	case models.Bool:
 		val, err := cast.ToBoolE(reading)
--- a/configs/datastorage/datastorage-device.yaml
+++ b/configs/datastorage/datastorage-device.yaml
@@ -31,3 +31,24 @@ deviceResources:
         { type: "Float64", readWrite: "RW" , mediaType : "text/plain" }
       units:
         { type: "String", readWrite: "R" }
+  - name: jpeg
+    description: "Jpeg image"
+    properties:
+      value:
+        { type: "Binary", readWrite: "RW" , mediaType : "image/jpeg" }
+      units:
+        { type: "String", readWrite: "R" }
+  - name: png
+    description: "png image"
+    properties:
+      value:
+        { type: "Binary", readWrite: "RW" , mediaType : "image/png" }
+      units:
+        { type: "String", readWrite: "R" }
+  - name: string
+    description: "String"
+    properties:
+      value:
+        { type: "String", readWrite: "RW" , mediaType : "text/plain"}
+      units:
+        { type: "String", readWrite: "R" }


internal/controller/storagemgr/storagedriver/storagehandler.go
```diff
@@ -196,6 +198,16 @@ func deviceHandler(writer http.ResponseWriter, request *http.Request) {
 	handler.processAsyncRequest(writer, request)
 }
 
+func convertToBase64(val []byte) string {
+	// Decode the CBOR format
+	var h codec.Handle = new(codec.CborHandle)
```
t25kim, line 203: Can you explain why we should do the CBOR processing before converting image byte streams to BASE64 directly?
Is there any problems without CBOR processing?

internal/controller/storagemgr/storagedriver/storagehandler.go
```diff
@@ -214,7 +226,10 @@ func (handler StorageHandler) newCommandValue(resourceName string, reading inter
 		if !ok {
 			return nil, fmt.Errorf(castError, resourceName, "not []byte")
 		}
-		result, errn = models.NewCommandValue(resourceName, timestamp, val, valueType)
+		Base64Val := convertToBase64(val)
```
t25kim, line 229: How about converting only image/jpeg and image/png `media types` to base64 rather than changing all binary data?

internal/controller/storagemgr/storagedriver/storagehandler_test.go
```diff
@@ -59,7 +59,7 @@ func TestNewCommandValue(t *testing.T) {
 		Type          models.ValueType
 		ErrorExpected bool
 	}{
-		{"Test A Binary", []byte{1, 0, 0, 1}, models.Binary, false},
+		{"Test A Binary", string("AA=="), models.String, false},
```
t25kim, line 62: I think it would be better to leave this line as the old one.

t25kim (commented):
Thanks for the awesome work!
Please take a look at my comments and share your idea.

internal/controller/storagemgr/storagedriver/storagehandler.go
sun-sharma: All binary data being transferred to EdgeX in in CBOR format, thus I had put a decoder. But you are right, I need to test it without the decoder too. I will do that and remove it if it is unnecessary.

internal/controller/storagemgr/storagedriver/storagehandler.go
sun-sharma: Sure, I think that can be a better approach. I will add a check for that.

internal/controller/storagemgr/storagedriver/storagehandler_test.go
sun-sharma: This was done as earlier all Binary data was being converted to Base64 thus this was necessary. Post adding a check I think this would not be necessary. I will remove it if check passes.

sun-sharma:
Thanks you @t25kim for the review. I will make the changes.

internal/controller/storagemgr/storagedriver/storagehandler.go
sun-sharma: It worked, I removed that part.

internal/controller/storagemgr/storagedriver/storagehandler.go
sun-sharma: Done.

internal/controller/storagemgr/storagedriver/storagehandler_test.go
sun-sharma: Post change, I reverted this.

MoonkiHong (approved):
Verified the required revisions w.r.t. the reviews from @t25kim. LGTM.

internal/controller/storagemgr/storagedriver/storagehandler.go
```diff
@@ -214,7 +221,14 @@ func (handler StorageHandler) newCommandValue(resourceName string, reading inter
 		if !ok {
 			return nil, fmt.Errorf(castError, resourceName, "not []byte")
 		}
-		result, errn = models.NewCommandValue(resourceName, timestamp, val, valueType)
+		if resourceName == "jpeg" || resourceName == "png" {
+			Base64Val := convertToBase64(val)
+			valueType = models.String
+			resourceName = "string"
```
t25kim, line 227: I think we don't need to change the resourceName.

t25kim (commented):
PTAL

internal/controller/storagemgr/storagedriver/storagehandler.go
sun-sharma: I tested it, It works. Thanks !
Name

t25kim (approved):
LGTM. Thanks!

tdrozdovsky (approved):
LGTM

Issue #306, opened 2021-05-31
[DataStorage] Initiation Scenario Change


t25kim:
**Is your feature request related to a problem? Please describe.**

At the moment, edge-orchestration initiates DataStorage if [configuration file is in](https://github.com/lf-edge/edge-home-orchestration-go/blob/de5d0aa9b4b300bbbf3dbe8ad1f244786d66fac2/internal/controller/storagemgr/storage.go#L49). This is static. 
I'd like to change the mechanism to make datastorage connect dynamically. 

**Describe the solution you'd like**

There are two types of edge-orchestration that run edgex and don't run edgex. 
## edge-orchestration running edgex on localhost
1. If edge-orchestration has an ENV value related to DataStorage, create configuration files with the host 127.0.0.1 for [Clients data and metadata](https://github.com/lf-edge/edge-home-orchestration-go/blob/de5d0aa9b4b300bbbf3dbe8ad1f244786d66fac2/configs/datastorage/configuration.toml#L22). 
2. edge-orchestration initiates the DataStorage and set DataStorage in the service list.

## edge-orchestration not running edgex on localhost
1. If edge-orchestration that does not start DataStorage finds DataStorage service from other edge-orchestration's mdnsTXT, it creates configuration files with the host running the DataStorage service for [Clients data and metadata](https://github.com/lf-edge/edge-home-orchestration-go/blob/de5d0aa9b4b300bbbf3dbe8ad1f244786d66fac2/configs/datastorage/configuration.toml#L22).
2. edge-orchestration starts the DataStorage.

PR #314, opened 2021-06-09
[DataStorage] Build configuration files dynamically

### LINKED ISSUE #302: [DataStorage] Configuration File Creation and Modification
### Problem
An user willing to use DataStorage feature should put configuration files in the `/var/edge-orchestration/datastorage/` folder in the present and it's difficult to dynamically modify configuration files such as [configuration.toml](https://github.com/lf-edge/edge-home-orchestration-go/blob/3e36d86ec3906872b84ba3cb17175b74c08b256a/configs/datastorage/configuration.toml) and [datastorage-device.yaml](https://github.com/lf-edge/edge-home-orchestration-go/blob/3e36d86ec3906872b84ba3cb17175b74c08b256a/configs/datastorage/datastorage-device.yaml). 

### Suggestion
How about creating or revising configuration files during the [StartStorage](https://github.com/lf-edge/edge-home-orchestration-go/blob/3e36d86ec3906872b84ba3cb17175b74c08b256a/internal/controller/storagemgr/storage.go#L48) function?

t25kim:
- Configuration files(configuration.toml and datastorage-device.yaml) are generated automatically
  if the SERVICE environment variable contains DataStorage
- Name in configuration files would be set to datastorage-{deviceID}

# Description

Fixes #302 
Related to #306 

## Type of change

# How Has This Been Tested?
1. Delete configuration files in /var/edge-orchestration/datastorage/
2. Run edge-orchestration with SERVICE environment variable.
```
$ docker run -it --rm --privileged --network="host" --name edge-orchestration -e SERVICE=DataStorage -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro edge-orchestration:coconut
```
3. Check the device profile from edgex-core-metadata from a web browser
  ex) http://localhost:48081/api/v1/deviceprofile

### RESULT 
image

**Test Configuration**:
* Hardware: x86-64

# Checklist:


MoonkiHong (approved):
+1. Thank you! 😄 

1 code smell result from SonarCloud is from the data type of storage in the configuration file and it is not the issue.

tdrozdovsky (approved):
LGTM

MoonkiHong:
@suresh-lc PTAL! 😄

internal/orchestrationapi/orchestration.go
```diff
@@ -201,7 +201,7 @@ func (o OrchestrationBuilder) Build() Orche {
 func (o *orcheImpl) Start(deviceIDPath string, platform string, executionType string) {
 	resourceMonitorImpl.StartMonitoringResource()
 	o.discoverIns.StartDiscovery(deviceIDPath, platform, executionType)
-	o.storageIns.StartStorage()
+	o.storageIns.StartStorage("")
```
suresh-lc, line 204: How is the handling of host being run in different machine taken care, since null is being passed from orchestration, wanted to check on this.

suresh-lc:
Since toml and yaml are being programmatically generated, how can a device manufacture add his own device configuration definition. Any thought process of supporting it in future. Or do we think we should only support the 6 pre-defined data types.

internal/orchestrationapi/orchestration.go
t25kim: Thanks for the review!

The host is the address of edgex-data and edgex-metadata, and DataStorage of edge-orchestration tries to connect to edgex with that address.
DataStorage in a edge-orchestration should not handle or control other edge-orchestrations.

Let me explain briefly about StartStorage() with the host, which is set to "".
1. StartStorage checks whether SERVICE environment variable contains the word, DataStorage.
2-1. If edge-orchestration has the environment variable, it generates configuration files with host "127.0.0.1".
2-2. If edge-orchestration doesn't have the environment variable, it checks the host address and generates configuration files when the host is not "".
3. Run DataStorage server with the configuration generated.
* If edge-orchestration doesn't have environment variable in terms of DataStorage and doesn't have proper host address, the edge-orchestration runs with the configuration files in the /var/edge-orchestration/datastorage/.

I'm not sure if I have answered your question sufficiently.
Please refer to the source code of StartStorage() for more information.

t25kim:
> Since toml and yaml are being programmatically generated, how can a device manufacture add his own device configuration definition. Any thought process of supporting it in future. 

As you pointed out, this PR is the initial source code and there is little way to configure it in terms of device manufacturers. It would be nice to implement source code to support your idea.

> Or do we think we should only support the 6 pre-defined data types.

Yes. This is where we are. We can extend supporting data types! Please contribute ideas and source code!

internal/orchestrationapi/orchestration.go
suresh-lc: Thanks for your kind reply, after referring to files had posted the query. Anyways thanks for the details.

suresh-lc:
> > Since toml and yaml are being programmatically generated, how can a device manufacture add his own device configuration definition. Any thought process of supporting it in future.
> 
> As you pointed out, this PR is the initial source code and there is little way to configure it in terms of device manufacturers. It would be nice to implement source code to support your idea.
> 
> > Or do we think we should only support the 6 pre-defined data types.
> 
> Yes. This is where we are. We can extend supporting data types! Please contribute ideas and source code!

Might be we capture these in the issue to say we start off with 6 data types. Sure will contribute towards the idea and code.

suresh-lc (approved):
Thanks for the proposed programmatic config file creations

Issue #324, opened 2021-07-01
[DataStorage] Upgrade to EdgeX v2.0 (Ireland)


t25kim:
We should get out of [temporary use](https://github.com/lf-edge/edge-home-orchestration-go/blob/34ad844318457cb2cbae5d359cfa1d6ec0998411/go.mod#L5) and need to check the possibility of official EdgeX v2.0 for Home Edge.

t25kim:
I made the `ds_ireland` branch to test DataStorage with edgex v2.0. 

1. Run EdgeX containers without security 
```
$ cd deployments/datastorage
$ docker-compose up -d
```
2. Save configuration files to `/var/edge-orchestration/datastorage/`
3. Create the container image
```
./build.sh container x86_64
```
4. Run edge-orchestration 
```
docker run -it -d --privileged --network="host" --name edge-orchestration -e EDGEX_SECURITY_SECRET_STORE=false -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro edge-orchestration:coconut
```

MoonkiHong:
> I made the `ds_ireland` branch to test DataStorage with edgex v2.0.
> 
> 1. Run EdgeX containers without security
> 
> ```
> $ cd deployments/datastorage
> $ docker-compose up -d
> ```
> 
> 1. Save configuration folders and files in [configs/datastorage folder](https://github.com/lf-edge/edge-home-orchestration-go/tree/ds_ireland/configs/datastorage) to `/var/edge-orchestration/datastorage/`
> 2. Create the container image
> 
> ```
> $ ./build.sh container x86_64
> ```
> 
> 1. Run edge-orchestration
> 
> ```
> $ docker run -it -d --privileged --network="host" --name edge-orchestration -e EDGEX_SECURITY_SECRET_STORE=false -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro edge-orchestration:coconut
> ```

@t25kim This is really a nice approach.

MoonkiHong:
@sun-sharma PTAL. (including #326) in terms of the test w.r.t. the PR.

MoonkiHong:
@tdrozdovsky PTAL in terms of build methodology from the impact of EdgeX v2.0 upgrade.

[...]

PR #326, opened 2021-07-05
[DO NOT MERGE] Import EdgeX v2.0.0

t25kim:
# Description

This is a temporary branch for testing edgex v2.0.0 against DataStorage.

WIP #324 

## Issue 
EdgeX uses 0mq which is a shared library and makes cross-compilation difficult.

## Type of change

# How Has This Been Tested?
See https://github.com/lf-edge/edge-home-orchestration-go/issues/324#issuecomment-873858913
```

### CODE REVIEW

**Comment by tdrozdovsky** [comment]
Thank you! I think the Go version should also be changed here:
```
docs/platforms/orange_pi3/orange_pi3.md:72:- go compiler (install a version not lower than 1.12.5)
docs/platforms/raspberry_pi3/raspberry_pi3.md:70:- go compiler (install a version not lower than 1.12.5, recommended v1.14.x)
docs/platforms/x86_64_linux/x86_64_linux.md:40:- go compiler
```
It is also possible to exclude the build possibility of edge-orchestration directly on boards (orange_pi3 and raspberry_pi3 ) Since it may be difficult to support Go newest version. What do you think about it?

File: deployments/datastorage/docker-compose.yml [hunk 1]
```diff
@@ -1,4 +1,4 @@
-#  * Copyright 2020 Intel Corporation.
```
tdrozdovsky, line 1: Are you sure you need to change the year of issue?

t25kim: I pasted the orignal [code](https://github.com/edgexfoundry/edgex-compose/blob/main/docker-compose-no-secty.yml#L1).

tdrozdovsky: Ok, Great!

t25kim:
> Thank you! I think the Go version should also be changed here:
> 
> ```
> docs/platforms/orange_pi3/orange_pi3.md:72:- go compiler (install a version not lower than 1.12.5)
> docs/platforms/raspberry_pi3/raspberry_pi3.md:70:- go compiler (install a version not lower than 1.12.5, recommended v1.14.x)
> docs/platforms/x86_64_linux/x86_64_linux.md:40:- go compiler
> ```

Thanks for the information!

> 
> It is also possible to exclude the build possibility of edge-orchestration directly on boards (orange_pi3 and raspberry_pi3 ) Since it may be difficult to support Go newest version. What do you think about it?

We build the edge-orchestration in terms of OS architecture rather than development boards, and I think we don't need to remove it at this time because golang supports different OS architectures. I'd like to talk about this issue when we encounter this problem.

t25kim:
EdgeX v2.1(Jakarta) has been released. We will think of what version of edgex should be taken.

PR #330, opened 2021-07-09
Added configuration of the basic docker image

tdrozdovsky:

# Description

Due to the lack of flexibility to use the basic image of the Doker container and its settings, I added the ability to configure it. Now users can use their docker file and settings in it. Also, after the adoption of this PR, we can contact [Vitalii55](https://github.com/Vitalii55/edge-home-orchestration-go) so that he adds his version to use the Alpine 3.7 image on x86_64.

P.S. I tested his Alpine 3.7 image and it works well, it should be noted that the size of the image is about 45Mb. This option will be available after applying the PR from [Vitalii55](https://github.com/Vitalii55/edge-home-orchestration-go/blob/add-alpine-linux-support/Dockerfile)

## Type of change

# How Has This Been Tested?
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration. Below is just an example and you can delete it when you contribute.

```
$ make distclean
$ make create_context CONFIGFILE=x86_64c
$ make 
$ make run
```

**Test Configuration**:
* Firmware version: (OS type and version, e.g., Ubuntu 18.04, etc.)
* Hardware: x86-64
* Toolchain: Recommended Docker and Go version.

# Checklist:


Kconfig
```diff
@@ -10,6 +10,13 @@ config CONFIGURATION_FILE_NAME
 	help
 		Configuration file name
 
+config DOCKER_FILE_NAME
```
t25kim, line 13: I think we can shorten the length of `CONFIGURATION_FILE_NAME` and `DOCKER_FILE_NAME` such as `CONFIGFILE` and `DOCKERFILE`. What do you think?

Kconfig
```diff
@@ -10,6 +10,13 @@ config CONFIGURATION_FILE_NAME
...
+	string "Dockerfile name"
+	default "ubuntu16df"
```
t25kim, line 15: For ease of maintenance, I'd like to use only the OS name, not the version. What do you think?
```suggestion
	default "ubuntu"
```

t25kim (commented):
Thanks for your contribution.
Please take a look at my comments!

Kconfig
tdrozdovsky: Ok, I'll do it!

Kconfig
tdrozdovsky: @t25kim I think that we must specify the version, as there may be other versions. For example, Ubuntu 18.04 or Alpine 3.7
Moreover, I think that we have to specify also sub-version for example Alpine 3.7: `alpine3.7` or `alpine3_7`. But these are only for the templates that are stored in repository, Users can save their with any names.

Tell me what do you think about it? And give examples, please!

Thank you!

t25kim: Thank you for sharing your thought. 
Your idea is the best if possible. However is it possible to maintain all versions with limited manpower? Just thinking of Ubuntu, there are several versions such as 16, 18, 20, and next one. Also how long will we support for each version? 
If we manage only one representative version for each OS, it is okay to me to attach a version. But maintaining all the versions doesn't make sense, and it's better to develop other features at that time.

tdrozdovsky: @t25kim I do not mind let's simplify. When it is necessary then and add.

tdrozdovsky: Done!

Kconfig
tdrozdovsky: Done!

t25kim (approved):
LGTM. Thanks!

MoonkiHong (approved):
+1! 😄

tdrozdovsky:
> 
> 
> +1! 😄

Thank you! When PR will be accepted, I will contact Vitalii55 to add its code for alpine image

t25kim:
> 
> Thank you! When PR will be accepted, I will contact Vitalii55 to add its code for alpine image

When it comes to Vitalii55's Dockerfile, does it build images in terms of arm and arm64?

tdrozdovsky:
> 
> 
> > Thank you! When PR will be accepted, I will contact Vitalii55 to add its code for alpine image
> 
> When it comes to Vitalii55's Dockerfile, does it build images in terms of arm and arm64?

I tested only x86_64 and Vitalii55 too. But I can check, and I will say the result. In any case, I think that it is necessary to accept code from an external developer, and then fix for arm if it does not work under arm

tdrozdovsky:
@t25kim I successfully compiled the alpine 3.12 image on Ubuntu 18.04 for `arm` and `arm64`, but haven't launched yet on board

MoonkiHong:
@suresh-lc PTAL. 😄

suresh-lc (approved):
LGTM

PR #368, opened 2021-08-19
Update the installation procedure of testing packages

tdrozdovsky:

# Description

Background:
When installing `gocov` and `gocov-html` packages with the `edge-home-orchestration-go ` folder, a description is added to the `go.mod` file. (go.mod is a tracked by the GitHub) The `go mod tidy` command makes check and optimize dependencies and since the packages are not included in the source code will be excluded from `go.mod`file. Therefore, `go.mod` file will always be marked as modified.

Update the installation procedure of `gocov` and `gocov-html`
 
## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/docs/platforms/x86_64_linux/x86_64_linux.md
+++ b/docs/platforms/x86_64_linux/x86_64_linux.md
@@ -46,6 +46,7 @@ Please see the below [How to work](#how-to-work) to know how to run Edge Orchest
 `$ export PATH=$PATH:$GOPATH/bin`
 
 - extra Go utilities: (optional)
+> Recommendation: Do not install the `gocov` and `gocov-html` utilities from the `edge-home-orchestration-go` folder.
   - [gocov](https://pkg.go.dev/github.com/axw/gocov)
   ```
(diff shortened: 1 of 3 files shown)


t25kim (commented):
How about adding the `go.mod` file to the .gitignore instead of recommendations?

tdrozdovsky:
If we exclude `go.mod` file from tracking, then after cloning, the repository will not have it. And we get a mistake like this:
`go: go.mod file not found in current directory or any parent directory; see 'go help modules'`

Or do you have other thoughts regarding this?

t25kim:
> 
> 
> If we exclude `go.mod` file from tracking, then after cloning, the repository will not have it. And we get a mistake like this:
> `go: go.mod file not found in current directory or any parent directory; see 'go help modules'`
> 
> Or do you have other thoughts regarding this?

I was mistaken. I think your suggestion is the best at the moment.

t25kim (approved):
LGTM. Thanks!

MoonkiHong (approved):
+1

PR #377, opened 2021-09-01
Add GetLogLevel() to Logmgr

t25kim:

# Description
This PR is for changing the logging level.
Logs are printed with the value applied to the `LOGLEVEL` environment variable. 
If the logging level is not set or incorrect, the `Info` level is set by default. 

## Type of change

# How Has This Been Tested?

Run Edge Orchestration with Warn logging level.
```
$ docker run -it -d --privileged --network="host" --name edge-orchestration -e LOGLEVEL=warn  -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro edge-orchestration:coconut
```

**Test Configuration**:
* Hardware: x86-64

# Checklist:

### DIFF
--- a/internal/common/logmgr/logmgr.go
+++ b/internal/common/logmgr/logmgr.go
@@ -48,7 +48,7 @@ func init() {
 			return function, filenline
 		},
 	}
-	logIns.Level = logrus.InfoLevel
+	logIns.Level = GetLogLevel()
 	logIns.Out = os.Stdout
 }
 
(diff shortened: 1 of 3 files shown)


internal/common/logmgr/logmgr.go
```diff
@@ -48,7 +48,7 @@ func init() {
 			return function, filenline
 		},
 	}
-	logIns.Level = logrus.InfoLevel
```
MoonkiHong, line 51: Thank you for your keen suggestion. This seems very useful. 💯 Overall it seems good to me, and I would like to lastly check out 1 point for my self clarification. 🙏 

If we deprecate `logrus.InfoLevel` and adopt a new feature by providing a new func `GetLogLevel`, do we also have to guide how developers use the customizing the log level to use it? Hope to educate me correctly. 😄

MoonkiHong (commented):
Overall, it seems good to me, so please educate me by checking out my comment below. 😁

internal/common/logmgr/logmgr.go
MoonkiHong: Thank you for your additional guidance! It really helps developers to properly use the optional log feature. 😄

MoonkiHong (approved):
LGTM. 💯

tdrozdovsky (approved):
LGTM

PR #434, opened 2021-12-09
Add initial prototype of Web UI

### LINKED ISSUE #425: Web UI
For the monitoring purpose, it must be good to have a web UI, such as [edgex foundry](https://github.com/edgexfoundry/edgex-ui-go#use-and-purpose) and [onos-gui](https://wiki.onosproject.org/display/ONOS/Web+UI+Tutorial+-+Creating+a+Topology+Overlay#WebUITutorialCreatingaTopologyOverlay-Overview).

The web UI can display the following items:
- Resource status
- List of service applications installed on the device
- List of devices in the home edge network

t25kim:

# Description
This PR is the first step to resolve #425. 
It runs the server for Web UI and displays simple text.

## Type of change

# How Has This Been Tested?
1. Run `Edge Orchestration` with `WEBUI` environment variable like below.
```
docker run -it -d --privileged --network="host" --name edge-orchestration -e WEBUI=true -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro lfedge/edge-home-orchestration-go:latest
```
2. Access the web page: http://localhost:49153
image

**Test Configuration**:
* OS type & version: Ubuntu 20.04
* Hardware: x86-64, arm
* Edge Orchestration Release: v1.1.1

# Checklist:

- [ ] I have performed a self-review of my own code


MoonkiHong (approved):
Validated from me. 💯 Thank you for this great contribution! 🙏

web/static/index.html
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE html><html lang="en"><head>
+  <meta charset="utf-8">
+  <title>Home Edge</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+<body>
+Home Edge Web UI
+</body></html>
```
NiranjanPatil25, line 7: </head> closing head tag is missing

NiranjanPatil25, line 7: Improper indentation followed, somewhere there are spaces and newlines, in some there isn't, need to fix

internal/webui/webui.go
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *******************************************************************************/
+
+package webui
+
+import (
+	"net/http"
+	"strconv"
+	"time"
+
```
NiranjanPatil25, line 24: remove unwanted empty lines

internal/webui/webui.go
```diff
@@ -0,0 +1,51 @@
...
+	uiPort      = 49153
+	log         = logmgr.GetInstance()
+)
+
+// Start starts the server for web UI
+func Start() {
+	s := &http.Server{
+		Handler:      uiHandler(),
+		Addr:         ":" + strconv.Itoa(uiPort),
+		WriteTimeout: 15 * time.Second,
```
NiranjanPatil25, line 39: Instead of having read/write timeout constant defined multiple places within code, defining it as constant variable at top or in environment variables can help in future maintainability

NiranjanPatil25 (approved):
Looks good

internal/webui/webui.go
t25kim: Thanks for the attentive review! Done.

web/static/index.html
t25kim: Done

t25kim: Done

internal/webui/webui.go
t25kim: Done

cmd/edge-orchestration/main.go
```diff
@@ -100,6 +101,7 @@ func orchestrationInit() error {
 
 	secure := os.Getenv("SECURE")
 	mnedc := os.Getenv("MNEDC")
+	ui := os.Getenv("WEBUI")
```
suresh-lc, line 104: In Home Edge there are many devices which are very low resource device. Say like motion sensor/Camera which doesnt have display on its own(within the device), having a web interface listing resources/services might not be used in that device. Hence how about making WebUI run as separate container(Process). And only devices with good resources(Display&Computing) can run the WebUI

suresh-lc:
> I think it is a good idea. In the case of using a simple page, we will not waste a lot of resources. If a situation arises with resource overruns, then it will be possible to transfer it to a separate container.
> 
> P.S. In any case, I think that the Web UI will be necessary for more convenient device configuration

What do you think of devices which don't have display unit self. In such devices Home Edge can run without WebUI, but other devices like TV/Mobile it can run with webUI.

tdrozdovsky:
> > I think it is a good idea. In the case of using a simple page, we will not waste a lot of resources. If a situation arises with resource overruns, then it will be possible to transfer it to a separate container.
> > P.S. In any case, I think that the Web UI will be necessary for more convenient device configuration
> 
> What do you think of devices which don't have display unit self. In such devices Home Edge can run without WebUI, but other devices like TV/Mobile it can run with webUI.

I think that the server does not need a display, Access can be obtained from the device on which it is

suresh-lc:
> > > I think it is a good idea. In the case of using a simple page, we will not waste a lot of resources. If a situation arises with resource overruns, then it will be possible to transfer it to a separate container.
> > > P.S. In any case, I think that the Web UI will be necessary for more convenient device configuration
> > 
> > 
> > What do you think of devices which don't have display unit self. In such devices Home Edge can run without WebUI, but other devices like TV/Mobile it can run with webUI.
> 
> I think that the server does not need a display, Access can be obtained from the device on which it is

WebUI to my understanding is going to refer to the various Dbs (to start off with and later can query system also) in Home Edge to display the resource or service info. So it can do so in any device which has display option is what I meant to explain.

suresh-lc:
Having the webui is nice to have feature. Just expressing my point as build within Home Edge or separate.

tdrozdovsky:
> Having the webui is nice to have feature. Just expressing my point as build within Home Edge or separate.

It would be nice if each device has a page for configuration, viewing the current status, etc. in this case Web UI should be within Home Edge.  

And I agree that for more complex queries (displaying the state of the entire network, etc.), a more powerful device will be needed.

cmd/edge-orchestration/main.go
t25kim: @suresh-lc @tdrozdovsky Thanks for the great idea! 
I like the idea that run containers separately. However bolt db for resource, service, etc in the Edge Orchestration container is not exposed. As a result, there must be a web ui server for the interfaces at Edge Orchestration container at the moment. We can run containers as your idea later, if we have any db container like MariaDB.

MoonkiHong:
@suresh-lc @tdrozdovsky This PR is to add a simple page for configure necessary aspects for the connected edge-home-orchestration based devices in the network. What about dealing with the suggestion from @suresh-lc in the latter and new PR?

Personally, I think that the both views are reasonable at this moment. 😁

t25kim:
@MoonkiHong @suresh-lc @tdrozdovsky Web UI is experimental at this stage. 
The back-end code (rest interface for db information) should be under the ` internal/webui/` folder and the front-end should be under the `web` folder.
Any devices can connect to the BE server to fetch information.

Issue #382, opened 2021-09-09
[Cloud Synchronization] Method to send/receive data with cloud infrastructure


suresh-lc:
**Is your feature request related to a problem? Please describe.**
Currently all the device generated data is stored using EdgeX container in the coredata db. With time the amount of data stored would be increasing exponentially. And this would in turn make the device slower for service execution due to lower memory. 

**Describe the solution you'd like**
In typical small devices(embedded system) the data would be stored at cloud infrastructure for future use. Hence it is always better to synchronize data to cloud. Method to save data to cloud infra is recommended.

**Describe alternatives you've considered**
Alternative method to save the data could be other high memory devices in the same network. But considering a Home ecosystem where all the devices are memory and CPU constrained.

**Additional context**
We need to define and consider clear method for the synchronization. Also the cloud system like AWS, Azure needs to be considered.

What/Where data resides : 
Within Home Edge we have the device meta data - like configuration, network and service info.
With EdgeX we have the device generated data being stored. 

Typical Use case:
Using Meta data - Understand the device configuration and services offered by these
Using Device data - Analytics on the device usage/performance, automation

Phase - I:
Synchronize the data residing in coredata to cloud system extending the functionality provided by EdgeX for cloud synchronization

Phase - II:
Method - I - Provide method to synchronize the data from coredata to cloud independent of EdgeX.
Method - II - Provide system to store the data generated by devices within Home Edge and consecutively method synchronize to cloud

MoonkiHong:
Good suggestion! 💯 @suresh-lc 

As a phase I, do you think of which candidate would be feasible for us to start with?

t25kim:
Thanks for the great suggestion!
As for Method - I in Phase - I, we don't have to think about it. coredata is tightly coupled with edgex so it can use edgex cloud sync mechanism.
When it comes to Method - II, it would be good to provide some interfaces not only the data generated by home edge but also any service applications which want to publish data to cloud.

suresh-lc:
> Thanks for the great suggestion! As for Method - I in Phase - II, we don't have to think about it. coredata is tightly coupled with edgex so it can use edgex cloud sync mechanism. When it comes to Method - II, it would be good to provide some interfaces not only the data generated by home edge but also any service applications which want to publish data to cloud.

Thanks for the suggestions. When coming to method II, the synch of data to cloud can be done via REST  or MQTT method. Supporting one method should be good to start of with, what is your opinion.

[...]

PR #446, opened 2021-12-20
[CloudSync] Adding the API to be exposed to thirdparty for Data Publish

nitu-s-gupta:
Signed-off-by: Nitu Gupta <nitu.gupta@samsung.com>

# Description
provide mechanism to expose API that can be used by third party apps to publish data to MQTT broker running on AWS endpoint

Fixes # (#382 )

## Type of change

# How Has This Been Tested?
```
```
3. The logs are checked using the command
```
docker logs -f edge-orchestration
```
4. From the another terminal/post make a curl command as follows to publish data using home edge to the broker running on AWS endpoint
```
> --header 'Content-Type: text/plain' \
> --data-raw '{
>     "appid": "TV1",
>     "payload": "{Another data from TV1}",
>     "topic": "home/livingroom",
>     "url" : "ec2-54-144-110-26.compute-1.amazonaws.com"
> }'
```
5. You should obtain the following message for successful publish
```
{"Message":"Data published successfully to Cloud","RemoteTargetInfo":"","ServiceName":""}
```

**Test Configuration**:
* Hardware: (x86-64)

# Checklist:

### DIFF
--- a/internal/restinterface/externalhandler/externalhandler.go
+++ b/internal/restinterface/externalhandler/externalhandler.go
@@ -26,7 +26,9 @@ import (
 
 	"github.com/lf-edge/edge-home-orchestration-go/internal/common/logmgr"
 
+	mqttmgr "github.com/lf-edge/edge-home-orchestration-go/internal/common/mqtt"
 	"github.com/lf-edge/edge-home-orchestration-go/internal/common/networkhelper"
+	"github.com/lf-edge/edge-home-orchestration-go/internal/controller/cloudsyncmgr"
 	"github.com/lf-edge/edge-home-orchestration-go/internal/controller/securemgr/verifier"
 	"github.com/lf-edge/edge-home-orchestration-go/internal/db/bolt/common"
 	"github.com/lf-edge/edge-home-orchestration-go/internal/orchestrationapi"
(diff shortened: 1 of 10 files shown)


internal/restinterface/externalhandler/externalhandler.go
```diff
@@ -351,6 +359,101 @@ SEND_RESP:
 	h.helper.Response(w, respEncryptBytes, http.StatusOK)
 }
 
+// APIV1RequestCloudSyncmgrPost handles cloudsync publish request from service application
+func (h *Handler) APIV1RequestCloudSyncmgrPost(w http.ResponseWriter, r *http.Request) {
```
t25kim, line 363: If I call this 10 times with the same host, will 10 mqtt client go routines be created?

t25kim (commented):
Thanks for the contribution. 
Please take a look at my comments.

internal/restinterface/externalhandler/externalhandler.go
```diff
@@ -351,6 +359,101 @@ SEND_RESP:
...
+		goto SEND_RESP
+	}
+	url, ok = appCommand["url"].(string)
+	if !ok {
+		responseMsg = orchestrationapi.InvalidParameter
+		goto SEND_RESP
+	}
+	cloudsyncmgr.GetInstance().StartCloudSync(url)
+	for mqttmgr.GetClient() == nil {
+		//Wait till the client is set
```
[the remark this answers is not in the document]
nitu-s-gupta: If using sleep what time value would be used. Its not always same in settting client

internal/restinterface/externalhandler/externalhandler.go
nitu-s-gupta: currently it is not storing the client data. Also one more thing when same host is sending the conenctn request the previous connection is closed and new is opened.

suresh-lc: Wanted to make the base code with single connection with the client. Going forward we think of scenarios where multiple service applications can call the send API to send data to different broker URL's. So in that case maintaining separate mqtt client instance for each service application would be better(and only one instance for each host). Use of db to maintain the same or so could be option. We plan to do this once the base code is merged as part of next steps.

tdrozdovsky (commented):
LGTM

tdrozdovsky (approved):

t25kim (approved):
LGTM Thanks!

suresh-lc (approved):
LGTM

MoonkiHong (approved):
+1

PR #452, opened 2022-01-07
[CloudSync] Saving the Client Data to avoid reconnection for same client in same home edge session

nitu-s-gupta:
Signed-off-by: Nitu Gupta <nitu.gupta@samsung.com>

# Description

Currently the client data was not stored and hence every time third party app request to publish the data connection is created.But this creates multiple connections by the same app to the broker. So this PR devises the method to store the client data for each app in a home edge session, so that app can directly publish the data.
The data of the client is stored in hashmap and app id sent in the request is used as client id to maintain uniqueness as required by mosquitto. When a client request for disconnect, the data is deleted from the table and on every connect successful the data is added to the table

Fixes # (#382)

## Type of change

# How Has This Been Tested?
```
```
3. From the another terminal/post make a curl command as follows to publish data using home edge to the broker running on AWS endpoint

```
--data-raw '{
    "appid": "com.nokia.homeedge1",
    "topic": "home1/livingroom",
    "url" : "ec2-54-205-12-98.compute-1.amazonaws.com"
}'

--data-raw '{
    "appid": "com.nokia.homeedge1",
    "payload": "{Data from same client again}",
    "topic": "home1/livingroom",
    "url" : "ec2-54-205-12-98.compute-1.amazonaws.com"
}'
```
4.You should obtain the following message for successful publish
```
{"Message":"Data published successfully to Cloud","RemoteTargetInfo":"","ServiceName":""}
```
5. The Broker logs are checked in AWS and we observe following logs
```
1641480800: New connection from 49.207.219.60 on port 1883.
1641480800: New client connected from 49.207.219.60 as com.nokia.homeedge1 (p2, c1, k30).
1641480800: No will message specified.
1641480800: Sending CONNACK to com.nokia.homeedge1 (0, 0)
1641480800: Received PUBLISH from com.nokia.homeedge1 (d0, q0, r1, m0, 'home1/livingroom', ... (80 bytes))
1641480830: Received PINGREQ from com.nokia.homeedge1
1641480830: Sending PINGRESP to com.nokia.homeedge1
1641480850: Received PUBLISH from com.nokia.homeedge1 (d0, q0, r1, m0, 'home1/livingroom', ... (73 bytes))
```
Same client sending data but connection is established only once.

**Test Configuration**:
* OS type & version: ( Ubuntu 20.04)
* Hardware: (x86-64)

# Checklist:

### DIFF
--- a/internal/common/mqtt/mqttconfig.go
+++ b/internal/common/mqtt/mqttconfig.go
@@ -20,19 +20,20 @@ package mqtt
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
 	MQTT "github.com/eclipse/paho.mqtt.golang"
 )
 
-const clientID = "TestHomeEdge"
 const mqttPort = 1883
 
 // Client is a wrapper on top of `MQTT.Client`
 type Client struct {
 	ID   string
 	Host string
+	URL  string
 	Port uint
 	Qos  byte
 	sync.RWMutex
@@ -47,21 +48,11 @@ type Message struct {
 }
 
 var (
-	mqttClient *Client
+	clientData map[string]*Client
 )
 
 // Config represents an attribute config setter for the `Client`.
 type Config func(*Client)
-
-//SetClient sets the client
-func SetClient(client *Client) {
-	mqttClient = client
-}
-
-//GetClient gets the client set
-func GetClient() *Client {
-	return mqttClient
-}
 
 // SetClientID sets the mqtt client id.
 func SetClientID(id string) Config {
@@ -84,9 +75,34 @@ func SetPort(port uint) Config {
 	}
 }
 
+// InitClientData creates an initialized hashmap
+func InitClientData() {
+	clientData = make(map[string]*Client)
+}
+
+// CheckifClientExist used to check if the client conn object exist
+func CheckifClientExist(clientID string) *Client {
+	client := clientData[clientID]
+	return client
+}
+
+// addClientData is used to add the client object based on client id
+func addClientData(client *Client, clientID string) {
+	clientData[clientID] = client
+}
+
 //SetBrokerURL returns the broker url for connection
 func (c *Client) SetBrokerURL(protocol string) string {
 	return fmt.Sprintf("%s://%s:%d", protocol, c.Host, c.Port)
+}
+
+func checkforConnection(brokerURL string, mqttClient *Client) int {
+	if mqttClient == nil {
+		return 0
+	}
+	log.Info(logPrefix, mqttClient.URL)
+	connURL := fmt.Sprintf("%s://%s:%d", "tcp", brokerURL, mqttPort)
+	return strings.Compare(connURL, mqttClient.URL)
 }
 
 // NewClient returns a configured `Client`. Is mandatory
@@ -98,16 +114,12 @@ func NewClient(configs ...Config) (*Client, error) {
 	for _, config := range configs {
 		config(client)
 	}
-
 	copts := MQTT.NewClientOptions()
 	copts.SetAutoReconnect(true)
+	copts.SetClientID(client.ID)
 	copts.SetMaxReconnectInterval(1 * time.Second)
 	copts.SetOnConnectHandler(client.onConnect())
-	copts.SetConnectionLostHandler(func(c MQTT.Client, err error) {
-		log.Warn(logPrefix, " disconnected, reason: "+err.Error())
-		mqttClient.Connect()
-	})
-
+	copts.SetConnectionLostHandler(client.onConnectionLost())
 	client.ClientOptions = copts
 
 	return client, nil
--- a/internal/common/mqtt/mqttconnection.go
+++ b/internal/common/mqtt/mqttconnection.go
@@ -56,6 +56,12 @@ func (client *Client) onConnect() MQTT.OnConnectHandler {
 	return nil
 }
 
+//onConnectionLost callback is called in case of disconnection from broker
+func (client *Client) onConnectionLost() MQTT.ConnectionLostHandler {
+	log.Warn(logPrefix, "Connection to the client is disconnected")
+	return nil
+}
+
 // IsConnected checks if the client is connected or not
 func (client *Client) IsConnected() bool {
 	if client.Client == nil {
@@ -67,33 +73,59 @@ func (client *Client) IsConnected() bool {
 
 // Disconnect disconnects the connection
 func (client *Client) Disconnect(quiesce uint) {
+	log.Info(logPrefix, "Disconnect requested from client")
 	if client.Client != nil {
 		client.Client.Disconnect(quiesce)
 	}
 }
 
-//StartMQTTClient is used to initiate the client and set the configuration
-func StartMQTTClient(brokerURL string) {
+// StartMQTTClient is used to initiate the client and set the configuration
+func StartMQTTClient(brokerURL string, clientID string) string {
+	log.Info(logPrefix, "Starting the MQTT Client")
+	//Check if the client connection exist
+	mqttClient := CheckifClientExist(clientID)
+	// Check if the connection exist for same url
+	ifConn := checkforConnection(brokerURL, mqttClient)
+	if mqttClient != nil && ifConn == 0 {
+		log.Info(logPrefix, "Connection Object exist", mqttClient)
+		if mqttClient.IsConnected() {
+			log.Info(logPrefix, "Client is already connected")
+			return ""
+		}
+		connectErr := mqttClient.Connect()
+		if connectErr != nil {
+			log.Warn(logPrefix, connectErr)
+			return connectErr.Error()
+		}
+
+	} else if mqttClient != nil && ifConn != 0 {
+		//close previous connection and remove from table
+		mqttClient.Disconnect(1)
+		delete(clientData, clientID)
+	}
 
 	clientConfig, err := NewClient(
 		SetHost(brokerURL),
 		SetPort(uint(mqttPort)),
+		SetClientID(clientID),
 	)
 	if err != nil {
 		log.Warn(logPrefix, err)
+		return err.Error()
 	}
 	clientConfig.ClientOptions.SetOnConnectHandler(clientConfig.onConnect())
 	URL := clientConfig.SetBrokerURL("tcp")
 	log.Info(logPrefix, " The broker is", URL)
+	clientConfig.URL = URL
 	clientConfig.ClientOptions.AddBroker(URL)
 
 	connectErr := clientConfig.Connect()
 	if connectErr != nil {
 		log.Warn(logPrefix, connectErr)
+		return connectErr.Error()
 	}
-
-	SetClient(clientConfig)
-
+	addClientData(clientConfig, clientID)
+	return ""
 }
 
 //Publish is used to publish the client data to the cloud
--- a/internal/common/mqtt/mqttconnection_test.go
+++ b/internal/common/mqtt/mqttconnection_test.go
@@ -20,13 +20,14 @@ import (
 	"testing"
 )
 
-//const Host = "broker.emqx.io"
-const Host = "ec2-54-175-241-64.compute-1.amazonaws.com"
+const Host = "broker.hivemq.com"
 const port = "1883"
 
 func TestStartMQTTClient(t *testing.T) {
 	t.Run("Success", func(t *testing.T) {
-		StartMQTTClient(Host)
+		InitClientData()
+		StartMQTTClient(Host, "testClient")
+		t.Log("Client connected")
 	})
 
 }
--- a/internal/controller/cloudsyncmgr/cloudsync.go
+++ b/internal/controller/cloudsyncmgr/cloudsync.go
@@ -21,6 +21,7 @@ package cloudsyncmgr
 import (
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/lf-edge/edge-home-orchestration-go/internal/common/logmgr"
 	mqttmgr "github.com/lf-edge/edge-home-orchestration-go/internal/common/mqtt"
@@ -33,9 +34,8 @@ const (
 // CloudSync is the interface for starting Cloud synchronization
 type CloudSync interface {
 	InitiateCloudSync(isCloudSet string) error
-	StartCloudSync(host string) error
 	//implemented by external REST API
-	RequestCloudSyncConf(message mqttmgr.Message, topic string, clientID string) string
+	RequestCloudSyncConf(host string, clientID string, message mqttmgr.Message, topic string) string
 }
 
 //CloudSyncImpl struct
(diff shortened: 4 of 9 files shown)


t25kim (commented):
Thanks for the contribution. 
Prior to test, I wonder if this PR has been tested with multiple app IDs and multiple URLs like below.
case 1: {id1, url1} > {id1, url2} > {id1, url1}
case 2: {id1, url1} > {id2, url1} > {id1, url2} > {id2, url2}
case 3: {id1, url1} > {id2, url2} > {id1, url1}
...

internal/common/mqtt/mqttconnection.go
```diff
@@ -93,6 +114,7 @@ func StartMQTTClient(brokerURL string) {
 	}
 
 	SetClient(clientConfig)
+	addClientData(clientConfig, clientID)
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/common/mqtt/mqttconnection_test.go
```diff
@@ -21,12 +21,14 @@ import (
 )
 
 //const Host = "broker.emqx.io"
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/common/mqtt/mqttconnection.go
```diff
@@ -67,14 +73,28 @@ func (client *Client) IsConnected() bool {
 
 // Disconnect disconnects the connection
 func (client *Client) Disconnect(quiesce uint) {
+	log.Info(logPrefix, "Disconnect requested from client")
 	if client.Client != nil {
 		client.Client.Disconnect(quiesce)
 	}
 }
 
 //StartMQTTClient is used to initiate the client and set the configuration
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/common/mqtt/mqttconfig.go
```diff
@@ -84,11 +91,42 @@ func SetPort(port uint) Config {
...
+	client := clientData[clientID]
+	if client == nil {
+		return false
+	}
+	mqttClient = client
+	return true
+
+}
+
+//addClientData is used to add the client object based on client id
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/common/mqtt/mqttconfig.go
```diff
@@ -84,11 +91,42 @@ func SetPort(port uint) Config {
 	}
 }
 
+//InitClientData creates an initialized hashmap
+func InitClientData() {
+	clientData = make(map[string]*Client)
+}
+
+//checkifClientExist used to check if the client conn object exist
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/common/mqtt/mqttconfig.go
```diff
@@ -64,10 +67,14 @@ func GetClient() *Client {
 }
 
 // SetClientID sets the mqtt client id.
-func SetClientID(id string) Config {
-	return func(c *Client) {
-		c.ID = id
-	}
+func SetClientID(id string) {
+	clientID = id
+
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/common/mqtt/mqttconfig.go
```diff
@@ -84,11 +91,42 @@ func SetPort(port uint) Config {
 	}
 }
 
+//InitClientData creates an initialized hashmap
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/common/mqtt/mqttconfig.go
```diff
@@ -84,11 +91,42 @@ func SetPort(port uint) Config {
...
+	return true
+
+}
+
+//addClientData is used to add the client object based on client id
+func addClientData(client *Client, clientID string) {
+	clientData[clientID] = client
+}
+
 //SetBrokerURL returns the broker url for connection
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!!

internal/common/mqtt/mqttconfig.go
```diff
@@ -48,6 +49,8 @@ type Message struct {
 
 var (
 	mqttClient *Client
+	clientID   string
```
t25kim, line 52: I think the `clientID` as a global variable is not a good approach. 
The `clientID` seems to be changed frequently by appID. I'm afraid a race condition problem when there are two requests from appID A and appID B at the same time.
I think you should use the `ID` of Client struct. What do you think?

internal/common/mqtt/mqttconnection.go
```diff
@@ -67,14 +73,27 @@ func (client *Client) IsConnected() bool {
...
 	if client.Client != nil {
 		client.Client.Disconnect(quiesce)
 	}
 }
 
-//StartMQTTClient is used to initiate the client and set the configuration
-func StartMQTTClient(brokerURL string) {
-
+// StartMQTTClient is used to initiate the client and set the configuration
+func StartMQTTClient(brokerURL string, clientID string) {
```
t25kim, line 83: I think the parameter for `brokerURL` is the IP or Domain name of host. 
It is confused with the result of [SetBrokerURL](https://github.com/lf-edge/edge-home-orchestration-go/blob/0589ede2610efbf9dd029d07a190ffe607d889d4/internal/common/mqtt/mqttconfig.go#L114). 
Please use other name instead of brokerURL to clarify the definition.

t25kim (commented):
PTAL!

internal/restinterface/externalhandler/externalhandler.go
```diff
@@ -431,11 +431,11 @@ func (h *Handler) APIV1RequestCloudSyncmgrPost(w http.ResponseWriter, r *http.Re
 		responseMsg = orchestrationapi.InvalidParameter
 		goto SEND_RESP
 	}
-	cloudsyncmgr.GetInstance().StartCloudSync(url)
+	cloudsyncmgr.GetInstance().StartCloudSync(url, appID)
```
t25kim, line 434: Can you define function names of cloudsyncmgr more properly?
`StartCloudSync` seems as if to run `cloudsyncmgr`, but it's not and the purpose of `StartCloudSync` at the moment is to run a mqtt client instead of cloudsyncmgr.
Also what about calling `mqttmgr.StartMQTTClient(host, clientID)` in the `RequestCloudSyncConf()`?

nitu-s-gupta: What name would you suggest? We had initially thought of using startmqtt but thought in cloudsync mgr package this name as main function would sound confusing and hence we went with StartCloudSyncmgr function.

t25kim: If it were me, I don't make `StartCloudSync`. Perhaps I call `mqttmgr.StartMQTTClient` in the `RequestCloudSync`. I also recommend you to use [channels](https://go.dev/blog/concurrency-timeouts) instead of [for loop](https://github.com/lf-edge/edge-home-orchestration-go/blob/0589ede2610efbf9dd029d07a190ffe607d889d4/internal/restinterface/externalhandler/externalhandler.go#L435).

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: Done!!

internal/restinterface/externalhandler/externalhandler.go
nitu-s-gupta: Moved StartMQTTClient to RequestCloudSyncConf and used wiatgroup instead of sleep to check if the client creation is completed instead of for lopp
Done!

nitu-s-gupta:
> Thanks for the contribution. Prior to test, I wonder if this PR has been tested with multiple app IDs and multiple URLs like below. case 1: {id1, url1} > {id1, url2} > {id1, url1} case 2: {id1, url1} > {id2, url1} > {id1, url2} > {id2, url2} case 3: {id1, url1} > {id2, url2} > {id1, url1} ...

Apart from this have also checked for the following scenarios
1. id1,url1 >id1,url2(broken)
2. id1,empty url
3. id1,broken url1
4. id1,url1>id1,url2>id2,url1>id1,broken url2

Please add more scenarios if required

internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -65,37 +65,50 @@ func (c *CloudSyncImpl) InitiateCloudSync(isCloudSet string) (err error) {
...
-		log.Info("Starting the CLoudsync Mgr")
-		go mqttmgr.StartMQTTClient(host)
-	}
-	return
-}
-
 // RequestCloudSyncConf is  configuration request handler
-func (c *CloudSyncImpl) RequestCloudSyncConf(message mqttmgr.Message, topic string, clientID string) string {
+func (c *CloudSyncImpl) RequestCloudSyncConf(host string, clientID string, message mqttmgr.Message, topic string) string {
 	log.Info(logPrefix, "Publishing the data to the cloud")
```
t25kim, line 78: What about checking `isCloudSyncSet` here? If `false`, return a message regarding `cloudsync is not active'.

t25kim (commented):
PTAL. Thanks!

internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -65,37 +65,50 @@ func (c *CloudSyncImpl) InitiateCloudSync(isCloudSet string) (err error) {
...
-	mqttClient = mqttmgr.GetClient()
-	mqttmgr.SetClientID(clientID)
 	resp := ""
-	if mqttClient.IsConnected() {
+	var wg sync.WaitGroup
```
t25kim, line 80: What about implementing a routing for publishing a message like [detectNetworkChgRoutine()](https://github.com/lf-edge/edge-home-orchestration-go/blob/565cc43ac637dcd22ffc0d9206584103632fb5a4/internal/controller/discoverymgr/discovery.go#L124) instead of `WaitGroup`?

nitu-s-gupta: This waitgroup is used instead of sleep to allow completion of go routine its not just related to network so for me it is correctly used. Thanks

t25kim: WaitGroup is synchronous. I recommend you to implement this asynchronously.

nitu-s-gupta: But in this case, making it asynchronous wwould not serve the purpose. In the scenario of timeout, sleep was giving an error because the wait time was less than the retry time of mosquitto.connect. In such a scenario empty message was given to the user. Hence a synchronous wwait grp is needed so that the connect result is obtained if error or success and then only further stmts are executed.

t25kim: Make sense. I thought this should be handled asynchronously and recommended you to implement it via goroutine and channel. If this should be run synchronously, it must be good to get rid of goroutine, waitgroup and channel. Sorry for your inconvenience.

internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -65,37 +65,50 @@ func (c *CloudSyncImpl) InitiateCloudSync(isCloudSet string) (err error) {
...
+		log.Info("Starting the Cloudsync Mgr")
+		if len(host) == 0 {
+			return "No broker host defined"
+		}
+		wg.Add(1)
+		errs := make(chan string, 1)
+		go func() {
+			errs <- mqttmgr.StartMQTTClient(host, clientID)
+			resp = <-errs
+			log.Info("The error message received is ", resp)
```
t25kim, line 91: Please display this log if it's not empty.

internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -65,37 +65,50 @@ func (c *CloudSyncImpl) InitiateCloudSync(isCloudSet string) (err error) {
...
-func (c *CloudSyncImpl) RequestCloudSyncConf(message mqttmgr.Message, topic string, clientID string) string {
+func (c *CloudSyncImpl) RequestCloudSyncConf(host string, clientID string, message mqttmgr.Message, topic string) string {
 	log.Info(logPrefix, "Publishing the data to the cloud")
-	mqttClient = mqttmgr.GetClient()
-	mqttmgr.SetClientID(clientID)
 	resp := ""
-	if mqttClient.IsConnected() {
+	var wg sync.WaitGroup
+	if isCloudSyncSet {
+		log.Info("Starting the Cloudsync Mgr")
```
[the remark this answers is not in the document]
nitu-s-gupta: This will be removed as it when i will be adding negative check for isCLoudsyncset

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: This wwas testing log and hence will be removed

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: go routine is necessary because else threads would be needed to allow the client to be connected.but go routines are cheaper than threads. Hence for each connection instead of thread a go routine is used.so it will be needed

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: Done!

internal/controller/cloudsyncmgr/cloudsync.go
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -65,37 +65,50 @@ func (c *CloudSyncImpl) InitiateCloudSync(isCloudSet string) (err error) {
...
 			resp = errMsg
 		} else {
-			resp = "Data published successfully to Cloud"
+			resp = "Data published successfully to Cloud" + mqttClient.URL
 		}
 	} else {
-		resp = "Client not connected to Broker URL"
+		resp = ""
 	}
-
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: Done!

internal/common/mqtt/mqttconfig.go
```diff
@@ -70,6 +72,11 @@ func SetClientID(id string) Config {
 	}
 }
 
+// GetClientID sets the mqtt client id.
+func GetClientID() string {
+	return mqttClient.ID
+}
+
```
t25kim, line 79: Nothing to call this.

internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -65,37 +65,49 @@ func (c *CloudSyncImpl) InitiateCloudSync(isCloudSet string) (err error) {
...
+		wg.Done()
+
+	}()
+	wg.Wait()
+	if resp != "" {
+		errresp := fmt.Sprintf("Error Connecting MQTT -> %s", resp)
+		return errresp
+	}
+
+	mqttClient = mqttmgr.GetClient()
```
t25kim, line 102: What about using a local variable for Client struct?
```suggestion
	mqttClient := mqttmgr.GetClient(clientID)
```

I think you can modify GetClient() to get Client struct from the `clientData` map.

internal/common/mqtt/mqttconfig.go
```diff
@@ -48,6 +49,7 @@ type Message struct {
 
 var (
 	mqttClient *Client
```
t25kim, line 51: I don't think global variable for a Client struct is a good way.

t25kim (commented):
PTAL

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: SInce added setter hence added the getter for future use. Will remove it for now

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: yes, mqttClient was used as shared variable in the initial design. Now with each connection having its own now its not required.

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: similar to above comment,

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: Done!!

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: The global variable is removed. All the client data is obtained from hashtable or nil value is returned

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: Done

t25kim (approved):
LGTM Thanks for your sincere efforts!

tdrozdovsky (approved):
LGTM

MoonkiHong (approved):
+1

PR #465, opened 2022-02-02
[CloudSync] Adding TLS to secure MQTT Client Broker Communication

nitu-s-gupta:
# Description

Certificates can be generated using openssl library and copied to /var/edge-orchestration/mqtt/certs folder. This will aloow the secure communication between mqtt broker and client.

Fixes # (#382)

## Type of change

# How Has This Been Tested?
1.MQTT mosquitto broker is configured to be running in the AWS endpoint. The certificates are copied to /etc/mosquitto/certs folder and corresponding path is specified in the configuration file
In the client side the root CA certificates are copied in the /var/edge-orchestration/mqtt/certs
```
```
3. From the another terminal/post make a curl command as follows to publish data using home edge to the broker running on AWS endpoint

```
--data-raw '{
    "appid": "com.nokia.homeedge1",
    "topic": "home1/livingroom",
    "url" : "ec2-54-205-12-98.compute-1.amazonaws.com"
}'
```

**Test Configuration**:
* Hardware: (x86)

# Checklist:


internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -29,6 +29,7 @@ import (
 
 const (
 	logPrefix = "[cloudsyncmgr]"
+	mqttPort  = 8883
```
t25kim, line 32: It would be better to set 8883 during `secure` mode. Otherwise, the number must be set to 1883.

t25kim (commented):
Thanks for the contribution! Please take a look at my comments.

internal/common/mqtt/mqttconfig.go
```diff
@@ -80,6 +87,15 @@ func InitClientData() {
 	clientData = make(map[string]*Client)
 }
 
+func (c *Client) setProtocol() {
+	port := c.Port
+	if port == 8883 {
```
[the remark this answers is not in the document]
nitu-s-gupta: Done!

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: Seems better option. Done!

internal/common/mqtt/mqttconfig.go
```diff
@@ -92,21 +107,36 @@ func addClientData(client *Client, clientID string) {
...
 		return 0
 	}
 	log.Info(logPrefix, mqttClient.URL)
-	connURL := fmt.Sprintf("%s://%s:%d", "tcp", brokerURL, mqttPort)
+	connURL := fmt.Sprintf("%s://%s:%d", mqttClient.protocol, brokerURL, mqttPort)
 	return strings.Compare(connURL, mqttClient.URL)
 }
 
+//NewTLSConfig creates a tls config for mqtt client
+func NewTLSConfig() *tls.Config {
```
t25kim, line 124: Please add a return for an error.
```suggestion
func NewTLSConfig() (*tls.Config, error) {
```

internal/common/mqtt/mqttconfig.go
```diff
@@ -92,21 +107,36 @@ func addClientData(client *Client, clientID string) {
...
+	}
+	certpool.AppendCertsFromPEM(ca)
+	return &tls.Config{
+		RootCAs: certpool,
+	}
+}
+
 // NewClient returns a configured `Client`. Is mandatory
 func NewClient(configs ...Config) (*Client, error) {
+	tlsconfig := NewTLSConfig()
```
t25kim, line 139: If `NewTLSConfig` returns an error, `NewClient` also should returns an error.

t25kim (commented):
PTAL!

internal/common/mqtt/mqttconfig.go
```diff
@@ -19,15 +19,21 @@
...
 	"sync"
 	"time"
 
 	MQTT "github.com/eclipse/paho.mqtt.golang"
 )
 
-const mqttPort = 1883
+const (
+	edgeDir      = "/var/edge-orchestration"
+	caCertConfig = edgeDir + "/mqtt/certs/cacert.pem"
```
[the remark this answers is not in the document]
nitu-s-gupta: This was kept under mqtt to keep it independent of other module.Hence mqtt info under mqtt folder. @tdrozdovsky please let me know if u have different thought

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: Done!!

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: Done!!

internal/common/mqtt/mqttconfig.go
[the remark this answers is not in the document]
tdrozdovsky: Good point! I think for now it might be here. In the future, I will try to automate the key deployment process and think about it more carefully.

t25kim (approved):
LGTM Thanks!

MoonkiHong (approved):
+1

internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -65,6 +67,13 @@ func (c *CloudSyncImpl) InitiateCloudSync(isCloudSet string) (err error) {
 		if strings.Compare(strings.ToLower(isCloudSet), "true") == 0 {
 			log.Println("CloudSync init set")
```
tdrozdovsky, line 68:
```suggestion
			log.Info(logPrefix, "CloudSync init set")
```

internal/controller/cloudsyncmgr/cloudsync.go
```diff
@@ -65,6 +67,13 @@ func (c *CloudSyncImpl) InitiateCloudSync(isCloudSet string) (err error) {
...
 			isCloudSyncSet = true
+			secure := os.Getenv("SECURE")
+			if len(secure) > 0 {
+				if strings.Compare(strings.ToLower(secure), "true") == 0 {
+					log.Println(logPrefix, "Orchestration init with secure option")
```
tdrozdovsky, line 73:
```suggestion
					log.Info(logPrefix, "Orchestration init with secure option")
```

tdrozdovsky, line 73: `logPrefix = "[cloudsyncmgr]"` change to `logPrefix = "[cloudsyncmgr] "`

internal/common/mqtt/mqttconnection.go
```diff
@@ -115,7 +115,8 @@ func StartMQTTClient(brokerURL string, clientID string) string {
 		return err.Error()
 	}
 	clientConfig.ClientOptions.SetOnConnectHandler(clientConfig.onConnect())
-	URL := clientConfig.SetBrokerURL("tcp")
+	clientConfig.setProtocol()
+	URL := clientConfig.SetBrokerURL()
 	log.Info(logPrefix, " The broker is", URL)
```
tdrozdovsky, line 120: Before `logPrefix = "[MQTTConnectionMgr]"` should be change to 1logPrefix = "[MQTTConnectionMgr] "`

```suggestion
	log.Info(logPrefix, "The broker is", URL)
```

internal/common/mqtt/mqttconfig.go
```diff
@@ -92,19 +108,33 @@ func addClientData(client *Client, clientID string) {
...
+	connURL := fmt.Sprintf("%s://%s:%d", mqttClient.protocol, brokerURL, mqttPort)
 	return strings.Compare(connURL, mqttClient.URL)
 }
 
+//NewTLSConfig creates a tls config for mqtt client
+func NewTLSConfig() (*tls.Config, error) {
+	certpool := x509.NewCertPool()
+	ca, err := ioutil.ReadFile(caCertConfig)
+	if err != nil {
+		log.Warn(err.Error())
```
tdrozdovsky, line 129:
```suggestion
		log.Warn(logPrefix, err.Error())
```

internal/common/mqtt/mqttconfig.go
```diff
@@ -120,7 +150,19 @@ func NewClient(configs ...Config) (*Client, error) {
...
 	copts.SetConnectionLostHandler(client.onConnectionLost())
+	secure := os.Getenv("SECURE")
+	if len(secure) > 0 {
+		if strings.Compare(strings.ToLower(secure), "true") == 0 {
+			tlsconfig, err := NewTLSConfig()
+			if err != nil {
+				return nil, err
+			}
+			copts.SetTLSConfig(tlsconfig)
+			// TODO Use Username and password to provide authorization to MQTT Broker zuncomment the below two lines to enable authorization after creating password file for each user
```
tdrozdovsky, line 161:
```suggestion
			// TODO: Use Username and Password to provide authorization to MQTT Broker uncomment the below two lines to enable authorization after creating password file for each user. This is temporary solution!
```

tdrozdovsky (changes requested):

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: what is the change in logPrefix? its already cloudsyncmgr

internal/common/mqtt/mqttconnection.go
nitu-s-gupta: Done!!

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: Done!!

nitu-s-gupta: Done!!

internal/controller/cloudsyncmgr/cloudsync.go
nitu-s-gupta: Done!!

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: This is not a temporary solution. Its the solution as suggested by mosquitto broker. So its user dependent and hence commented.

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: Done!

internal/common/mqtt/mqttconfig.go
tdrozdovsky: 1) login and password cannot be stored in the program and must be exported from a secure storage 
2) more suitable option is a signed certificate with a fixed SAN (we had difficulties with this, but I will investigate this issue in more detail later) 
So I think this is a temporary solution.

nitu-s-gupta: Yes , it should be ideally but you have to pass the password through copts only thats the architecture design of Mosquitto broker . Secondly Certificates through SAN can give you secure communication channel which is already done. The issue was resolved by me. There is no issue in it. So I think there is some confusion from your side. SAN certificates were fine and are working fine now. As already mentioned in previous PR also the issue was resolved and it wasnot certificates issue.

suresh-lc (approved):
+1

internal/common/mqtt/mqttconfig.go
tdrozdovsky: As I wrote above, this is a temporary solution, since the password cannot be stored in the program.

tdrozdovsky: The use of a fixed SAN certificate provides authentication without passing a password. (And while me and you have not been able to run this option)

nitu-s-gupta: > As I wrote above, this is a temporary solution, since the password cannot be stored in the program.

Agree to this point of yours. But still the password has to be passed as string be it readin from any file as per the mosquitto design requirement. So i think its reading a pssword file in string ad passing a string. Now, since its user dependent, its upto the user how he wants to read it from file, or memory etc. I think this discussion can be taken in separate thread since this will not impact the current lines of code. Hence this can be approved

tdrozdovsky: FYI: a secure channel can be established with any user, including a hacker.

nitu-s-gupta: > The use of a fixed SAN certificate provides authentication without passing a password. (And while me and you have not been able to run this option)

No, I was able to run with SAN certificate and password. So I have tried all three modes of security that the mosquitto broker provides and al have been running suceessfully without error

nitu-s-gupta: > FYI: a secure channel can be established with any user, including a hacker.

Yes, because its role is just to provide the secure channel. Thats where the password comes into picture. Which is way of providing ACL(Acess Control List) list of users who can acess. And With SAN i have added password for twwo users and only those two users could access. 
To clear the air of Confusion following are the different modes I have tried and tested sucessfully
1. No security:: Any user can access the broker specifying the url
2. Secure Channel: Any User can access the broker providing the SAN certificates
3. Secure Channel with password :: Only specific users can access the broker after prviding the password and SAN certificates.

t25kim: Sorry. I had not reviewed this carefully. What about receiving Username and Password as parameters of REST API? I think commenting them is not a good approach.

tdrozdovsky: Let's clear situation:
**After accepting this PR, we will not be able to release, not even a v1.1.7**. Since the container downloaded from Dockerhub will not be able to work with the broker. Where in the documentation and source code will this be mentioned?

nitu-s-gupta: I have commented it because its user choice that what level of security user wants. And If i uncomment it it will be always secure mode and will be checking the broker with password. Receiving username password via REST API, would make it more vulnerable I think, not sure about it,

nitu-s-gupta: > Let's clear situation: **After accepting this PR, we will not be able to release, not even a v1.1.7**. Since the container downloaded from Dockerhub will not be able to work with the broker. Where in the documentation and source code will this be mentioned?

Documentation its already mentioned. SOurce code //TODO line states clearly that it has to be uncommented after generating a password file for particular user

tdrozdovsky: At a minimum, we need to put this data into an external file and edge-orchestration read them from file

nitu-s-gupta: I m not denying to put password in the file and read it. Its not about how you read the password. Thats not the issue here.
I have just mentioned no matter how you read it be it API or file but in the end it has to be passed as string in copts only

t25kim: > I have commented it because its user choice that what level of security user wants. And If i uncomment it it will be always secure mode and will be checking the broker with password. Receiving username password via REST API, would make it more vulnerable I think, not sure about it,

The problem of static input is that the code is not working if there are two brokers with different password and username on AWS.

nitu-s-gupta: I think how the password is read can be discussed in separate thread. Passing password thorugh copts so that broker recognises and receives it is limitation of mosquitto broker. Sepaarate PR can be raised for reading of password with better secure way after discussing with our security expert @tdrozdovsky :)
@Taras please let me know if you agree

nitu-s-gupta: > > I have commented it because its user choice that what level of security user wants. And If i uncomment it it will be always secure mode and will be checking the broker with password. Receiving username password via REST API, would make it more vulnerable I think, not sure about it,
> 
> The problem of static input is that the code is not working if there are two brokers with different password and username on AWS.

The code is commented. So static input is not coming into picture. Keeping it static was purposefully to give user idea that string has to be passed. If this is creating so much confusion this commented lines can be removed altogether. But then only documentation wont give clear picture to a naive user

t25kim: @tdrozdovsky Security is too difficult... I have another question about the certificate. There is only one certificate in this PR. If brokers are different, I think they might want different certificates. Can the service applications send the cert information to edge orchestration via REST API?

tdrozdovsky: We can take it step by step
1) while we consider the option of communication with only one broker
2) we have two options: a) add the phrase that this is a temporary solution, accept this request, and urgently prepare to read the login and password from an external file; b) we finish reading the login and password from an external file or from another place

If accepted in the current version: we need remove the OpenSSF "passing" and "silver" badge from our project (because it violates security requirements ), and also the container that is downloaded from dockerhub will not work fully.

tdrozdovsky: I'm more concerned about the unprotected storage password in the code (hardcoded). And as I said earlier, we can implement the issues of secure interaction with the broker later

suresh-lc: We need to fix the secure location for password file storage . We can discuss that separately. And this file reading should not take much time. So once we fix the file location we can raise PR for that. @nitu-s-gupta : Please remove the lines..

t25kim: @nitu-s-gupta Please create an issue to discuss about cert and password.

tdrozdovsky (approved):
LGTM

internal/common/mqtt/mqttconfig.go
suresh-lc: > @nitu-s-gupta Please create an issue to discuss about cert and password.

https://github.com/lf-edge/edge-home-orchestration-go/issues/471

PR #498, opened 2022-02-23
Ignore docker pkg from dependabot since serious consideration is required

t25kim:

# Description

Packages related to docker engine should be excluded from dependabot since it requires [containerexecutor](https://github.com/lf-edge/edge-home-orchestration-go/blob/master/internal/controller/servicemgr/executor/containerexecutor/containerexecutor.go) code change.

## Type of change

- [x] CI system update

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,9 @@ updates:
     # Check for updates once a month
     schedule:
       interval: "monthly"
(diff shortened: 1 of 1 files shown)


tdrozdovsky (approved):
LGTM

MoonkiHong (approved):
+1

PR #532, opened 2022-03-23
[CloudSync] Save the Subscribed data & Adding API to get data

nitu-s-gupta:
# Description
After Subscribing to the topic, Code has been added to store the data published on subscribed topic and to provide the API to retrieve the stored data
UsageExample

Getting the data subscribed to a topic

curl --location --request GET 'http://{EdgeOrchestration IP}:56001/api/v1/orchestration/cloudsyncmgr/getsubscribedata/<topicname>/<appid>'

Fixes # (#382)

## Type of change

# How Has This Been Tested?
```
```
3.Call the subscribe API using curl request

```
curl --location --request POST 'http://192.168.1.107:56001/api/v1/orchestration/cloudsyncmgr/subscribe' \
--data-raw '{
    "appid": "app2",
    "topic": "kitchen",
    "url" : "ec2-52-6-247-126.compute-1.amazonaws.com"
}'
```
4. Call publish on the same topic using curl request
```
curl --location --request POST 'http://192.168.1.107:56001/api/v1/orchestration/cloudsyncmgr/publish' \
--data-raw '{
    "appid": "app4",
    "topic": "kitchen",
    "url" : "ec2-52-6-247-126.compute-1.amazonaws.com",
    "payload" : "Test message for kitchen1..."
}'
```
5. Call the fetch data API using curl request for the topic kitchen and appid app2
```
curl --location --request GET 'http://192.168.1.107:56001/api/v1/orchestration/cloudsyncmgr/getsubscribedata/kitchen/app2'
```
You should get the message sent by app4 in above example. So the output received is
```
{"Message":"Test message for kitchen1..."}
```
6. Now Call the curl command for topic bedroom
```
curl --location --request GET 'http://192.168.1.107:56001/api/v1/orchestration/cloudsyncmgr/getsubscribedata/bedroom/app2'
```
The response received is 
```
{"Message":"Client is not subscribed to the topic bedroom"}
```

**Test Configuration**:
* Hardware: (x86-64)

# Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/docs/cloudsync_mqtt.md
+++ b/docs/cloudsync_mqtt.md
@@ -266,7 +266,8 @@ The edge orchestration is build and run using following command with option CLOU
 ```
 docker run -it -d --privileged --network="host" --name edge-orchestration -e CLOUD_SYNC=true -v /var/edge-orchestration/:/var/edge-orchestration/:rw -v /var/run/docker.sock:/var/run/docker.sock:rw -v /proc/:/process/:ro  lfedge/edge-home-orchestration-go:latest
 ```
-From the another terminal/post make a curl command as follows to publish data using home edge to the broker running on AWS endpoint
+From another terminal/post, it is recommended to make a curl command as follows to publish data using home edge to the broker running on AWS endpoint<br>
+***Please Note :***  Donot use '/' in topic name field.
 
 ```
 curl --location --request POST 'http://<ip where edgeorchestration is running>:56001/api/v1/orchestration/cloudsyncmgr/publish' \
@@ -274,7 +275,23 @@ curl --location --request POST 'http://<ip where edgeorchestration is running>:5
 --data-raw '{
     "appid": "<appid of service app>",
     "payload": "{Another data from TV1 and testdata}",
-    "topic": "home1/livingroom",
+    "topic": "home1-livingroom",
     "url" : "<AWS public IP>"
 }'
+```
+From another terminal/post,it is recommended to make a curl post command as follows to subscribe to topic using home edge to the broker running on AWS endpoint
+
+```
+curl --location --request POST 'http://<ip where edge-orchestration is running>:56001/api/v1/orchestration/cloudsyncmgr/subscribe' \
+--header 'Content-Type: text/plain' \
+--data-raw '{
+    "appid": "<appid of the service app>",
+    "topic": "kitchen1",
+    "url" : "<AWS public IP>"
+}'
+```
+From another terminal/post,it is recommended make a curl get command as follows to get the data received on subscribed topic by passing the clientid(appid) and topic in the request header
+
+```
+curl --location --request GET 'http://<ip where edge-orchestration is running>:56001/api/v1/orchestration/cloudsyncmgr/getsubscribedata/<topicname>/<appid>'
 ```
--- a/internal/common/mqtt/mqttconfig.go
+++ b/internal/common/mqtt/mqttconfig.go
@@ -49,14 +49,10 @@ type Client struct {
 	protocol string
 }
 
-//Message is used to wrap the app id and payload into one and publish to broker
-type Message struct {
-	AppID   string
-	Payload string
-}
-
 var (
-	clientData map[string]*Client
+	publishData     map[string]*Client
+	subscribeData   map[string]string    //topic-->published data
+	subscribeClient map[string][]*Client //topic-->multiple clients
 )
 
 // Config represents an attribute config setter for the `Client`.
@@ -83,9 +79,11 @@ func SetPort(port uint) Config {
 	}
 }
 
-// InitClientData creates an initialized hashmap
-func InitClientData() {
-	clientData = make(map[string]*Client)
+// InitMQTTData creates an initialized hashmap
+func InitMQTTData() {
+	publishData = make(map[string]*Client)
+	subscribeClient = make(map[string][]*Client)
+	subscribeData = make(map[string]string)
 }
 
 func (c *Client) setProtocol() {
@@ -98,13 +96,34 @@ func (c *Client) setProtocol() {
 
 // CheckifClientExist used to check if the client conn object exist
 func CheckifClientExist(clientID string) *Client {
-	client := clientData[clientID]
+	client := publishData[clientID]
 	return client
 }
 
-// addClientData is used to add the client object based on client id
-func addClientData(client *Client, clientID string) {
-	clientData[clientID] = client
+// addpublishData is used to add the client object based on client id
+func addPublishData(client *Client, clientID string) {
+	publishData[clientID] = client
+}
+
+// addSubscribeClient is used to add the client info for a topic
+func addSubscribeClient(client *Client, topic string) {
+	subscribeClient[topic] = append(subscribeClient[topic], client)
+}
+
+func addSubscribedData(topic string, message string) {
+	subscribeData[topic] = message
+}
+
+//GetSubscribedData is used to get the data for topic subscribed
+func GetSubscribedData(topic string, clientID string) string {
+	list := subscribeClient[topic]
+	for _, client := range list {
+		if client.ID == clientID {
+			return subscribeData[topic]
+		}
+	}
+	return "Client is not subscribed to the topic " + topic
+
 }
 
 //SetBrokerURL returns the broker url for connection
--- a/internal/common/mqtt/mqttconnection.go
+++ b/internal/common/mqtt/mqttconnection.go
@@ -19,7 +19,6 @@
 package mqtt
 
 import (
-	"encoding/json"
 	"time"
 
 	MQTT "github.com/eclipse/paho.mqtt.golang"
@@ -102,7 +101,7 @@ func StartMQTTClient(brokerURL string, clientID string, mqttPort uint) string {
 	} else if mqttClient != nil && ifConn != 0 {
 		//close previous connection and remove from table
 		mqttClient.Disconnect(1)
-		delete(clientData, clientID)
+		delete(publishData, clientID)
 	}
 
 	clientConfig, err := NewClient(
@@ -126,24 +125,19 @@ func StartMQTTClient(brokerURL string, clientID string, mqttPort uint) string {
 		log.Warn(logPrefix, connectErr)
 		return connectErr.Error()
 	}
-	addClientData(clientConfig, clientID)
+	addPublishData(clientConfig, clientID)
 	return ""
 }
 
 //Publish is used to publish the client data to the cloud
-func (client *Client) Publish(message Message, topic string) error {
+func (client *Client) Publish(message string, topic string) error {
 
 	log.Info(logPrefix, "Publishing the data to cloud")
-	payload, err := json.Marshal(message)
-	if err != nil {
-		log.Warn(logPrefix, "Error in Json Marshalling", err)
-		return err
-	}
 	mqttClient := client.Client
 	for mqttClient == nil {
 		time.Sleep(time.Second * 2)
 	}
-	token := mqttClient.Publish(topic, 0, true, payload)
+	token := mqttClient.Publish(topic, 0, true, message)
 	if token.Wait() && token.Error() != nil {
 		return token.Error()
 	}
@@ -154,6 +148,7 @@ func (client *Client) Publish(message Message, topic string) error {
 var messageHandler MQTT.MessageHandler = func(client MQTT.Client, msg MQTT.Message) {
 
 	log.Info(logPrefix, "Topic ", msg.Topic(), " registered.. ", string(msg.Payload()), " is the payload")
+	addSubscribedData(msg.Topic(), string(msg.Payload()))
 }
 
 //Subscribe is used to subscribe to a topic
@@ -168,5 +163,6 @@ func (client *Client) Subscribe(topic string) error {
 		return token.Error()
 	}
 	time.Sleep(time.Second)
+	addSubscribeClient(client, topic)
 	return nil
 }
--- a/internal/common/mqtt/mqttconnection_test.go
+++ b/internal/common/mqtt/mqttconnection_test.go
@@ -28,16 +28,16 @@ var port uint = 1883
 const InvalidHost = "invalid"
 
 func initializeTest(Host string, clientID string) {
-	InitClientData()
+	InitMQTTData()
 	StartMQTTClient(Host, clientID, port)
 }
 
 func TestStartMQTTClient(t *testing.T) {
 	t.Run("Success", func(t *testing.T) {
 		initializeTest(Host, "testClient")
-		client := clientData["testClient"]
+		client := publishData["testClient"]
 		if client != nil {
-			message := Message{"appid", "Test data for testing"}
+			message := "Test data for testing"
 			client.Publish(message, "home/livingroom")
 		}
 		err := StartMQTTClient(Host, "testClient", port)
@@ -47,7 +47,7 @@ func TestStartMQTTClient(t *testing.T) {
 		}
 	})
 	t.Run("Fail", func(t *testing.T) {
-		InitClientData()
+		InitMQTTData()
 		err := StartMQTTClient(InvalidHost, "testClientFailure", port)
 		expected := "dial tcp: lookup invalid: Temporary failure in name resolution"
 		if !strings.Contains(err, expected) {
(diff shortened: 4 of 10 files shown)


internal/common/mqtt/mqttconfig.go
```diff
@@ -49,14 +49,10 @@ type Client struct {
...
-type Message struct {
-	AppID   string
-	Payload string
-}
-
 var (
-	clientData map[string]*Client
+	publishData     map[string]*Client
+	subscribeData   map[string]string    //topic-->published data
+	subscribeClient map[string][]*Client //topic-->multiple clients
```
t25kim, line 55: What about not considering `appid`? I think it's very inefficient if a lot of different clients sends messages to same topic.
I also think we can merge these variables into one.

nitu-s-gupta: I hve followed the design where any number of clients can publish message on any topic and so appid is not used. But only those clients who have subscribed to the topic should receive the message. 
Keeping the vairable separate is to decouple the information, as the subscribeClient can hold the data for client and ill be used to delete the client wwhen the client is disconnceted so client object is required.Only appid will not help.
The subscribeData is used to store the topics mapped to the messages and this has no relation with client object. So to decouple the information they are kept separate.

t25kim: Assume that there are 3 applications that want to publish messages about `home` topic. In this case, current code create 3 `Client` variables for the same topic and broker will send same message to the 3 `Client` variable.

> I hve followed the design where any number of clients can publish message on any topic and so appid is not used. But only those clients who have subscribed to the topic should receive the message. 

I think this case we can control `Client` by `topic` instead of `clientID` and `Client` has a list to manage `clientID`.

> Keeping the vairable separate is to decouple the information, as the subscribeClient can hold the data for client and ill be used to delete the client wwhen the client is disconnceted so client object is required.Only appid will not help. The subscribeData is used to store the topics mapped to the messages and this has no relation with client object. So to decouple the information they are kept separate.

I disagree with the information should be decoupled. If the `Client` is managed by `topic`, it can hold the message instead of `subscribeData`. Also the `Client` can be removed if there is no item in the `clientID` list.

nitu-s-gupta: No, client cant be managed by topic, Mosquitto requires unique client id and topic can be repetative and so. If the same client is publishing data on same topic or diff topic, he will not be added 3 times, since that client is already present and so same connection object shall be used and only publishing shall happen. Since the data is stored topic wise, if the client published 3 message on home only the latest shall be present and not three times

t25kim: What about combining `topic` and `device iID` to create a `client ID` for broker?
If we look at the `edge-orchestration` as a client, it's very inefficient to create multiple objects on the same topic.
For example, Edge Orchestration is subscribing the topic 'home' with 'app1', 'app2', and 'app3'
If `app` publishes a message, Edge Orchestration receives same 3 messages like log below,
```
INFO[2022-03-24T07:52:59Z]mqttconnection.go:136 Publish [MQTTConnectionMgr] Publishing the data to cloud
INFO[2022-03-24T07:52:59Z]mqttconnection.go:156 func1 [MQTTConnectionMgr] Topic home registered.. {"AppID":"app","Payload":"test"} is the payload
INFO[2022-03-24T07:52:59Z]mqttconnection.go:156 func1 [MQTTConnectionMgr] Topic home registered.. {"AppID":"app","Payload":"test"} is the payload
INFO[2022-03-24T07:52:59Z]mqttconnection.go:156 func1 [MQTTConnectionMgr] Topic home registered.. {"AppID":"app","Payload":"test"} is the payload
```

internal/common/mqtt/mqttconfig.go
```diff
@@ -49,14 +49,10 @@ type Client struct {
...
 
-//Message is used to wrap the app id and payload into one and publish to broker
-type Message struct {
-	AppID   string
-	Payload string
-}
-
 var (
-	clientData map[string]*Client
+	publishData     map[string]*Client
```
t25kim, line 53: How about just set `publishData` as Client? I think we can publish various topic messages with a Client variable if we set `client ID` of broker to `device ID` of Edge Orchestration.

```suggestion
	publishData     Client
```

internal/common/mqtt/mqttconfig.go
t25kim: In terms of broker, we can use a combination of 'topic' and 'device id'.
In terms of edge orchestration, we use a combination of 'topic' and 'host' for hash.

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: If we make the clientID as deviceID of edgeorchestration, how will the message from different clients be distinguished? Another issue is then there will be only one connection of HomeEdge and not the Service Application since multiple connections from same ClientID(which in our case will be DeviceID) and hence connection will be again set and reset for any new client coming. 
PublishData is a HashMap storing the Client Object with clientID as key for ease and fast retrieval of client. Else the entire hashmap has to be scanned to find any client which will increase the time taken

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: > In terms of broker, we can use a combination of 'topic' and 'device id'

No, Because Topic,deviceId can be repeatative as deviceID will be same for all clients and multiple clients canhave same topics. For example HomeEdge DeviceId is 12345 same for any client, now client1 wants to register/publish on topic kitchen and client 2 also on kitchen So both clients will have kitchen,12345 so uniquness not present

internal/common/mqtt/mqttconfig.go
t25kim: > If we make the clientID as deviceID of edgeorchestration, how will the message from different clients be distinguished? 

Is it necessary to make multiple connections to the broker when we distinguish which client's message it is? Instead of making multiple connections to the same topic, the client mark who sent it on the message.

>Another issue is then there will be only one connection of HomeEdge and not the Service Application since multiple connections from same ClientID(which in our case will be DeviceID) and hence connection will be again set and reset for any new client coming. 

Make sense. It looks good to have hash map structure. However I would like to manage `publishData` through the combination of `broker host` and `topic`.

internal/common/mqtt/mqttconfig.go
t25kim: > > In terms of broker, we can use a combination of 'topic' and 'device id'
> 
> No, Because Topic,deviceId can be repeatative as deviceID will be same for all clients and multiple clients canhave same topics. For example HomeEdge DeviceId is 12345 same for any client, now client1 wants to register/publish on topic kitchen and client 2 also on kitchen So both clients will have kitchen,12345 so uniquness not present

Yes. That's my opinion. In terms of broker, we don't need to distinguish connections by client id.

nitu-s-gupta: So, u mean all the third party apps sending connection request, shall create only one connection, with client id deviceId,topic
So going forward, when the client calls disconnects which connection shall be terminated by homeedge?
Also then logically there will be one connection between broker and all client, And all clients shall be virtually added to hashmap on publish/subscribe method and removed on disconnect call.
Correct?
If that is the case then only deviceId is sufficient ,since broker needs only one variable for uniqueness and not combination f values.

internal/common/mqtt/mqttconfig.go
nitu-s-gupta: The connections were made not to diferentiate between the messages but between the clients. Broker takes just two vales, clientid and payload. So we cannot send more than two data and even if json is sent it will be taken as payload data. So if the same deviceID is sent as string, client distinction is not possible. So the design considered sending the appid of the application, that is requesting as client id to maintain uniqueness plus multiple apps of same client can request for different things. Hence is the design consideration in the current way

t25kim:
@nitu-s-gupta This is what I suggest. Can you draw a diagram for the ease of understanding?

image

internal/common/mqtt/mqttconfig.go
t25kim: > So, u mean all the third party apps sending connection request, shall create only one connection, with client id deviceId,topic So going forward, when the client calls disconnects which connection shall be terminated by homeedge?

Assume that edge-orchestration can have a requester list for host `broker.com` and topic `home`. 
There were 3 client id (app1, app2, app3) requested. If `app1` calls disconnection, Then edge-orchestration deletes `app1` from the requester list. If there is no client id in the requester list, we can terminate the connection between broker and edge-orchestration.  

> Also then logically there will be one connection between broker and all client, And all clients shall be virtually added to hashmap on publish/subscribe method and removed on disconnect call. Correct? If that is the case then only deviceId is sufficient ,since broker needs only one variable for uniqueness and not combination f values.

All client connects to edge-orchestration, not broker. Also I don't understand why we should subscribe same topic to same broker with multiple client ID because the data from broker with a topic is same. Subscribed data is not different by client ID.

nitu-s-gupta:
IMG20220329125156__01

t25kim:
> IMG20220329125156__01

Thanks for sharing the picture! Each of `Url hashmap`, `Publish Data` and `Subscription Info` blocks have multiple `Client` objects according to url and topic. Right? 
Assume that App2, App3, App4 have subscribed to topic `kitchen` and App1 publishes a message about topic `kitchen` and payload `Test Message`.
I think there are 3 `Client`, <`kitchen`, `App1` >, <`kitchen`, `App2`>, <`kitchen`, `App3`>, and each of them will receive the same `Test Message`.  It is more efficient to get message with a Client, <`kitchen`> in terms of memory and network usage. 

image

t25kim:
Sorry for my mistake. I've tested that we don't need to combine `device ID` and `topic` for the `Client ID` of broker.
I would like to change my suggestion like below. 
The `Client` of Edge Orchestration is managed by `broker url` and the `Client ID` of broker is managed by `Device ID` of Edge Orchestration.

image

nitu-s-gupta:
**URL hashmap** shall only have the list of client is so its a hashmap of <string>[]string to have a count on how many clients have been connected to a url. Say if a client calls disconnect going forwward when that api is made available to them, then only it will be removed from list. Now say if the list becomes empty which means none clients connected to the url in that case entry will be deleted.
**PublishData** again is of the type <string>string and shall hold latest posted message against topic. given a topic, message can be retrieved.
**SubscriptionInfo** is of the type <string>[]string which shall hold the list of clientids subscribed for the topic. 
In all of the three we are not storing the client object.

Assume that App2, App3, App4 have subscribed to topic kitchen and App1 publishes a message about topic kitchen and payload Test Message.
So in that case initially one entry for kitchen shall be made and App2,App3,App4 shall be added to the list i.e. Kitchen->App2->App3->App4
This entry is needed to prevent any app who has not subscribed to topic from viewing it. Then App1 wwhen published the data one entry is updated with topic kitchen in PublishData
Now say App5 request for topic kitchen, So the Subscription Info shall be searched and then it is observed that App5 is not registered to the topic and hence error message of not registered is sent back

t25kim:
> **URL hashmap** shall only have the list of client is so its a hashmap of []string to have a count on how many clients have been connected to a url. Say if a client calls disconnect going forwward when that api is made available to them, then only it will be removed from list. Now say if the list becomes empty which means none clients connected to the url in that case entry will be deleted. 

Make sense.

> **PublishData** again is of the type string and shall hold latest posted message against topic. given a topic, message can be retrieved. 

I think it should hold latest message against topic and `url` because messages between broker1's topic and broker2's topic might be different.

> **SubscriptionInfo** is of the type []string which shall hold the list of clientids subscribed for the topic. In all of the three we are not storing the client object.

It also needs to consider `url` information.
 
> Assume that App2, App3, App4 have subscribed to topic kitchen and App1 publishes a message about topic kitchen and payload Test Message. So in that case initially one entry for kitchen shall be made and App2,App3,App4 shall be added to the list i.e. Kitchen->App2->App3->App4 This entry is needed to prevent any app who has not subscribed to topic from viewing it. Then App1 wwhen published the data one entry is updated with topic kitchen in PublishData Now say App5 request for topic kitchen, So the Subscription Info shall be searched and then it is observed that App5 is not registered to the topic and hence error message of not registered is sent back

This flow seems great!

nitu-s-gupta:
> > **URL hashmap** shall only have the list of client is so its a hashmap of []string to have a count on how many clients have been connected to a url. Say if a client calls disconnect going forwward when that api is made available to them, then only it will be removed from list. Now say if the list becomes empty which means none clients connected to the url in that case entry will be deleted.
> 
> Make sense.
> 
> > **PublishData** again is of the type string and shall hold latest posted message against topic. given a topic, message can be retrieved.
> 
> I think it should hold latest message against topic and `url` because messages between broker1's topic and broker2's topic might be different.

Makes sense <<topic,url>string becomes published data
> 
> > **SubscriptionInfo** is of the type []string which shall hold the list of clientids subscribed for the topic. In all of the three we are not storing the client object.
> 
> It also needs to consider `url` information.
SusbcriptionInfo also becomes <topic,url>Clientid
> 
> > Assume that App2, App3, App4 have subscribed to topic kitchen and App1 publishes a message about topic kitchen and payload Test Message. So in that case initially one entry for kitchen shall be made and App2,App3,App4 shall be added to the list i.e. Kitchen->App2->App3->App4 This entry is needed to prevent any app who has not subscribed to topic from viewing it. Then App1 wwhen published the data one entry is updated with topic kitchen in PublishData Now say App5 request for topic kitchen, So the Subscription Info shall be searched and then it is observed that App5 is not registered to the topic and hence error message of not registered is sent back
> 
> This flow seems great!

nitu-s-gupta:
Can we accept this raised PR as initial complete working model for CloudSync and a separate PR with the design discussed and finalised can be raised as next PR since the design changes are for all the data structure, I would suggest completing the functionality in this PR and rasing new PR for design

t25kim (approved):
Thank you very much for your great contribution! 
I approved this as an [initial working model](https://github.com/lf-edge/edge-home-orchestration-go/pull/532#issuecomment-1082689932). PRs will be coming as a new design.

suresh-lc (approved):
thanks for the contribution. Further design enhancements can be taken as new PR

docs/cloudsync_mqtt.md
```diff
@@ -266,15 +266,32 @@ The edge orchestration is build and run using following command with option CLOU
 ```
 ```
-From the another terminal/post make a curl command as follows to publish data using home edge to the broker running on AWS endpoint
+From the another terminal/post make a curl command as follows to publish data using home edge to the broker running on AWS endpoint<br>
```
MoonkiHong, line 269:
```suggestion
From another terminal/post, it is recommended to make a curl command as follows to publish data using home edge to the broker running on AWS endpoint<br>
```

docs/cloudsync_mqtt.md
```diff
@@ -266,15 +266,32 @@ The edge orchestration is build and run using following command with option CLOU
...
     "appid": "<appid of service app>",
     "payload": "{Another data from TV1 and testdata}",
-    "topic": "home1/livingroom",
+    "topic": "home1-livingroom",
     "url" : "<AWS public IP>"
 }'
-```
\ No newline at end of file
+```
+From the another terminal make a curl post command as follows to subscribe to topic using home edge to the broker running on AWS endpoint
```
MoonkiHong, line 282:
```suggestion
From another terminal, it is recommended to make a curl post command as follows to subscribe to topic using home edge to the broker running on AWS endpoint
```

docs/cloudsync_mqtt.md
```diff
@@ -266,15 +266,32 @@ The edge orchestration is build and run using following command with option CLOU
...
+```
+curl --location --request POST 'http://<ip where edge-orchestration is running>:56001/api/v1/orchestration/cloudsyncmgr/subscribe' \
+--header 'Content-Type: text/plain' \
+--data-raw '{
+    "appid": "<appid of the service app>",
+    "topic": "kitchen1",
+    "url" : "<AWS public IP>"
+}'
+```
+From the another terminal make a curl get command as follows to get the data received on subscribed topic by passing the clientid(appid) and topic in the request header
```
MoonkiHong, line 293:
```suggestion
From another terminal, it is recommended to make a curl get command as follows to get the data received on subscribed topic by passing the clientid(appid) and topic in the request header
```

MoonkiHong (commented):
Please take a look at my comments. Overall it looks good and I left several comments on the documentation to guide the amended API and functionality for the better understanding of developers.

docs/cloudsync_mqtt.md
nitu-s-gupta: Done!!

docs/cloudsync_mqtt.md
nitu-s-gupta: Done!!

docs/cloudsync_mqtt.md
nitu-s-gupta: Done!!

tdrozdovsky (approved):

